PR #11, opened 2023-06-07
Add navigation

aliokan:
Done: 
- add main navigation
- remove navigation in home
- fix sessions.session render

## Before

image

image

## After

image

image


snenenenenenene:
I would maybe move the new navigation component below the AuContentHeader?

aliokan:
> 

@snenenenenenene I don't want to move the `<NavigationBar />` in all route templates. 

What do you prefer :
- move `<ContentHeader />` in application.hbs between `<AppHeader />` and `<NavigationBar />`  
- or move `<NavigationBar />` in content-header.hbs  ?

FYI, without design I took inspiration from : 
- https://appuniversum.github.io/ember-appuniversum/?path=/story/templates-app-with-breadcrumbs--page
- https://www.vlaanderen.be/bouwen-wonen-en-energie/elektriciteit-en-aardgas

snenenenenenene (approved):
@aliokan I just tried both options and reckon we should probably consult saska about it, i'm approving it for now and then we can improve on the design later! :)

PR #12, opened 2023-06-09
Real municipality data in municipality page + QoL fixes

snenenenenenene:
- Adding loading and error routes as sub-routes in order to not break the immersion of the app
- Adding real data to the municipality page
- Giving the user more info about the sidebar inputs


app/components/content-header.hbs
```diff
@@ -1,7 +1,11 @@
-<AuContentHeader @titlePartOne="Vlaanderen" @titlePartTwo="is linked data">
-  <img
-    sizes="50vw"
-    src={{"/assets/loket-header-1600.jpg"}}
-    alt="Foto van een laptop met daarop het vlaanderen logo."
-  />
-</AuContentHeader>
+<div style="height: 15rem; min-height: 15rem;">
```
aliokan, line 1: @snenenenenenene **suggestion (non-blocking)** Move inline css in `/styles` folder. 

Inline CSS is bad practice and should only be used as a last resort. I open a followup ticket to plan CSS cleaning: https://binnenland.atlassian.net/browse/BNB-171

aliokan (approved):
@snenenenenenene It's a huge PR, I didn't went in detail. Let's merge it to move forward

PR #16, opened 2023-06-15
Sessions // add infinite scroll

aliokan:
The following changes have been made in order to load more data when the user reaches the bottom of the session list in the `sessions/index` route:
- add `scroll` event in app config `customEvents` to be able to listen the event in components. 
- create a new component (InfiniteList)
   - it's a container that listens for the `scroll` event on itself and dispatches `loadMore` when the scroll reaches over 80%.
- add `sessions/index` controller to catches `loadMore`, then load and add extra sessions on model.
- prevent model reset in `sessions` route when we navigate through the app.
- add `currentPage` in sessions model

[Screencast from 15-06-23 10:12:37.webm](https://github.com/lblod/frontend-burgernabije-besluitendatabank/assets/3050307/284f7bbe-11e7-4a6a-a143-2b23cfd6b299)

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -7,6 +7,9 @@ export default class App extends Application {
   modulePrefix = config.modulePrefix;
   podModulePrefix = config.podModulePrefix;
   Resolver = Resolver;
+  customEvents = {
+    scroll: "scroll",
+  }
 }
 
 loadInitializers(App, config.modulePrefix);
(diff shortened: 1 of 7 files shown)


snenenenenenene (approved):
Everything looks great, I'll also work on implementing your InfiniteList component in the agendaItems list!

PR #18, opened 2023-06-16
feat: :sparkles: updated detail page to match the figma demo

snenenenenenene:
DETAIL PAGE
- added session display to the bottom of the detail page
- added an indicator on whether or not an agenda item has been decided
- added styling that looks like the figma document
- added a resolution ember data model to look if agenda item has been decided
- removed contact link on section to the right of the page


app/controllers/home.ts
```diff
@@ -4,11 +4,39 @@ import { action } from "@ember/object";
...
+  @tracked errorMsg = "";
+
+  @action
+  async loadMore() {
+    if (this.model && !this.isLoadingMore) {
+      this.isLoadingMore = true;
+      const nextPage = this.model.currentPage + 1;
+      const agendaItems = await this.store.query(
+        "agenda-item",
+        this.model.getQuery({ page: nextPage })
```
aliokan, line 27: Shouldn't we add the filters here?

tests/unit/utils/get-formatted-date-range-test.js
```diff
@@ -1,23 +0,0 @@
-import { module, test } from 'qunit';
-import { getFormattedDateRange } from 'frontend-burgernabije-besluitendatabank/utils/get-formatted-date-range';
-
-module('Unit | Utility | get-formatted-date-range', function () {
```
aliokan, line 4: Why did you remove these tests?

tests/unit/utils/get-formatted-date-test.js
```diff
@@ -1,17 +0,0 @@
-import { module, test } from 'qunit';
-import { getFormattedDate } from 'frontend-burgernabije-besluitendatabank/utils/get-formatted-date';
-
-module('Unit | Utility | get-formatted-date', function () {
```
aliokan, line 4: Same here

aliokan (changes requested):
I add some comments

app/controllers/home.ts
snenenenenenene: I thought filtering should be agnostic to paginaton so I added it all to the route

tests/unit/utils/get-formatted-date-range-test.js
snenenenenenene: I heard echoes of jodi telling us testing is for the testing team

app/templates/home.hbs
```diff
@@ -1,50 +1,51 @@
...
+        @placeholder={{"Alle gemeentes"}}
+        @noMatchesMessage={{"Geen gemeente gevonden"}}
+        @searchMessage="Aan het zoeken..."
+        @searchEnabled={{true}}
+        class="au-u-margin-right-small"
+        as |singleSelect|
+      >
+        {{singleSelect.label}}
+      </PowerSelect>
+      <AuButton
```
aliokan, line 41: The button has to be disabled when municipality is empty, or at least do not add empty filter in url `gemeentes=`. 
image

aliokan (changes requested):

app/controllers/home.ts
aliokan: I agree with you pagination doesn't need to know about filter, but with your current implementation the loading nextPage doesn't apply filters. You can inject filters in `this.model.getQuery` from the route.

app/routes/home.ts
```diff
@@ -49,14 +127,59 @@ export default class HomeRoute extends Route {
...
+  async model(params: any, transition: Transition<unknown>) {
+    // const model: any = this.modelFor("home");
+    // if (
+    //   model?.agendaItems?.toArray().length > 0 &&
+    // ) {
+    //   return model;
+    // }
+
+    const currentPage = 0;
+    const agendaItems = await this.store.query(
```
aliokan, line 156: We do not need to load agenda items to display the new Home page. 
image

app/controllers/home.ts
snenenenenenene: Fixed!

app/templates/home.hbs
```diff
@@ -12,7 +12,7 @@
   style="display: flex; justify-content: center; align-items: center;"
 >
 
-  <AuFormRow class="au-u-2-5">
```
aliokan, line 15: Is `au-u-2-5` not working?

tests/integration/components/vote-table-entry-test.js
```diff
@@ -0,0 +1,26 @@
+import { module, test } from 'qunit';
```
aliokan, line 1: that test can be removed ;)

aliokan (changes requested):
left 2 small comments

PR #21, opened 2023-06-22
Feat(municipality): Add municipality-list Service

aliokan:
In order to keep municipalities data in one place, migrate municipality-list data management in `MunicipalityList` service instead of routes. 

- add MunicipalityList service
- use it in home + agenda-items
- sort data alphabetically. 

image


snenenenenenene (commented):
Everything looks a lot better! 
However, I've noticed it sorts them, but also not completely (?)
Aalst for instance is not visible in the beginning, but I can look it up.
image

aliokan:
@snenenenenenene Good catch! The error is coming from the server answer. I'm investigating, shouldn't block this PR.

snenenenenenene (approved):
Let's go!

PR #22, opened 2023-06-23
Creation & implementation of Filter Sidebar

Denperidge:
- [x] Allow queryParam getting & setting without extra route/controller code needed
- [x] Update into agenda-items
- [x] Implement into sessions

PR #26, opened 2023-06-23
SCSS Refactoring

Denperidge:
What this PR DOES:
- [x] Refactor app.scss into multiple files
- [x] Move all inline styling into their own scss files
- [x] Use prefixes on route/component selectors and document why (because app.scss is one big file)

What this pull request DOESN'T DO:
- Fix search-sidebar/filter-sidebar styling. This branch is based on master, which doesn't yet have #22 . Dependant on whether this PR or #22 gets merged first, the filter-sidebar should be checked for styling and split off if needed
- In depth checks for redundant/duplicate styling. While I merged and checked what I could spot during the refactoring, it is possible that two routes have similar styling things that could be merged together in a custom class down the road


Denperidge:
TODO: use appuniverse where possible

aliokan (approved):
Look good! Thanks :pray:

PR #29, opened 2023-06-26
Fix detail page

aliokan:
## Fix geen verslag warning title

Before

image

After

image

##  Hide votes when vote is missing

Before

image

After

image

## Fix missing bottom margin on large window

Before

image

After

image


aliokan:
I merge it. To be part of tomorrow's demo

PR #30, opened 2023-06-27
Fix decision card link

aliokan:
Add route on Card header link in order to fix behavior when open a new window or tab. 

Fix made on `decision-card` component, `sessions.index` and `sessions.session` route template. 

- Before: `http://localhost:4200/agendapunten#`
- After: `http://localhost:4200/agendapunt/a9576440-fb47-11ed-9059-d3f31ceac058`


aliokan:
I merge it. To be part of tomorrow's demo

PR #40, opened 2023-07-05
Filter changes

Denperidge:
The changes include - but aren't limited to - moving the keyword filter in agenda-items, and a bunch of refactors


aliokan (approved):
looks good! :+1:

PR #41, opened 2023-07-05
Hide Load more button if no more data is available

Denperidge:
Implemented through letting the loadMore actions return a boolean! It feels like a bit of a weird solution but it works so hey

### DIFF
--- a/app/components/infinite-list.hbs
+++ b/app/components/infinite-list.hbs
@@ -9,7 +9,9 @@
     {{#if @isLoading}}
       <AuLoader />
     {{else}}
-      <AuButton {{action "loadMore"}}>Laad meer</AuButton>
+      {{#if @moreData}}
+        <AuButton {{action "loadMore"}}>Laad meer</AuButton>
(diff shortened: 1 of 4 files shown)


aliokan (commented):
**Question:** What do you think to use total amount of items provided by the server, instead the current workaround? 
A PR is ready to add the missing information : https://github.com/lblod/app-burgernabije-besluitendatabank/pull/14

Denperidge:
That works I guess! I'll wait with merging

Denperidge:
Paused until https://github.com/lblod/app-burgernabije-besluitendatabank/pull/14 merge

Denperidge:
@aliokan I was under the assumption that I would wait until the app pr merge, but do you perhaps prefer to first try @Windvis ' approach?

aliokan:
@Denperidge Both solutions looks good to me. https://github.com/lblod/app-burgernabije-besluitendatabank/pull/14 is merged and deployed

PR #43, opened 2023-07-06
Run linters and tests in CI

Windvis:
Lots of changes because of the Prettier format, but checking commit per commit (and skipping the formatting one) should be ok.


.eslintrc.js
```diff
@@ -0,0 +1,60 @@
...
+    {
+      // ts files
+      files: ['**/*.ts'],
+      extends: [
+        'plugin:@typescript-eslint/eslint-recommended',
+        'plugin:@typescript-eslint/recommended',
+      ],
+      rules: {
+        '@typescript-eslint/ban-ts-comment': 'warn',
+        '@typescript-eslint/ban-types': 'warn',
```
Windvis, line 29: These were causing a lot of errors so I changed them to warnings for now.

.prettierrc
Windvis, line 1: This file was different from the default one that's shipped with Ember so I reverted it to that one. There is no real reason to go with a different config.

.woodpecker/.verify-pr.yml
```diff
@@ -0,0 +1,19 @@
+pipeline:
+  install:
+    image: danlynn/ember-cli:4.7.0-node_16.17
```
Windvis, line 3: Not the same ember-cli version, but that's the only one with node 16 and ember-cli will use the version from the node_modules folder anyways.

app/controllers/map.ts
```diff
@@ -35,15 +35,6 @@ export default class MapComponent extends Controller {
 
   @tracked path: any = geoPath().projection(this.projection);
 
-  @computed(
```
Windvis, line 38: Nooooo idea why these were there. Computeds are old school stuff and never used on methods 😄.

app/controllers/municipality.ts
```diff
@@ -10,7 +10,6 @@ export default class MunicipalityController extends Controller {
 
   //@tracked options = ["antwerpen"];
   @tracked selected = '';
-  @action handleKeywordChange() {}
```
Windvis, line 13: Didn't seem to be used.

app/controllers/sessions/index.ts
```diff
@@ -27,17 +27,17 @@ export default class SessionsIndexController extends Controller {
...
+      const plannedStartMax = String(this.plannedStartMax) || undefined;
+      const municipalities = String(this.municipality) || undefined;
 
       const sessions = await this.store.query(
         'session',
         this.model.getQuery(
           nextPage,
-          (plannedStartMin = plannedStartMin),
-          (plannedStartMax = plannedStartMax),
-          (municipalities = municipalities)
```
Windvis: Another very funky thing. Not sure what was going on here 😄. eslint complained about the self-assignment as well.

ember-cli-build.js
```diff
@@ -37,9 +37,6 @@ module.exports = function (defaults) {
     '@appuniversum/ember-appuniversum': {
       disableWormholeElement: true,
     },
-    'ember-cli-babel': {
-      includePolyfill: true,
-    },
```
Windvis, line 42: Whoops, committed this by accident.

.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  overrides: [
+    {
+      files: '*.{js,ts}',
+      options: {
+        singleQuote: true,
```
Windvis, line 8: I can also remove this for now and stick to `"`. Not the default in Ember apps but it would introduce more noise here. I don't really have a preference. As long as it's consistent I'm happy 😄.

Denperidge (approved):
Are there any guidelines within LBLOD/ABB about styling/linting and/or which images/ember-cli to use during testing? Or is it an every repo for themselves thing?

If the latter (or the first and those guidelines have been applied/followed), looks good!

Windvis:
> Are there any guidelines within LBLOD/ABB about styling/linting and/or which images/ember-cli to use during testing? Or is it an every repo for themselves thing?

Not really. Ideally we would also use the `madnificent/ember` image in CI, but it doesn't include a browser so we can't use it to run the tests. So we just need to match the node versions as good as possible.

PR #44, opened 2023-07-06
Use root level pipeline conditions

Windvis:
This has the benefit that it won't spawn pipelines with only a clone step if the condition isn't matched.


Denperidge (approved):
Looks good!

PR #45, opened 2023-07-07
Remove unneeded config from the ember-cli-build.js file

Windvis:


ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-          global: true,
-        },
-      },
-      useSwcParser: true,
-      resolve: {
-        extensions: ['.js', '.json'],
-      },
-    },
-    hinting: IS_TEST,
-    tests: IS_TEST,
```
Windvis, line 24: This disables the `/tests` route which is very handy.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-        node: {
-          global: true,
-        },
-      },
-      useSwcParser: true,
-      resolve: {
-        extensions: ['.js', '.json'],
-      },
-    },
-    hinting: IS_TEST,
```
Windvis, line 23: Not sure what this does exactly.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-    tests: IS_TEST,
-    'ember-cli-babel': {
-      includePolyfill: IS_PROD,
-    },
```
Windvis, line 27: I don't think the polyfill is needed? Assuming we only support modern browsers?

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-    autoprefixer: {
-      sourcemap: false,
-    },
```
Windvis, line 30: Not sure why this was needed. If we don't want the autoprefixing we should just remove the ember-cli-autoprefixer addon.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-    sourcemaps: {
-      enabled: IS_PROD,
-    },
```
Windvis, line 33: It seems strange to disable source maps in development builds so I would either always enable them, or keep the default behaviour (enabled in dev, not in prod).

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-    fingerprint: {
-      enabled: true,
-    },
```
Windvis, line 36: We don't need this in development.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
     '@appuniversum/ember-appuniversum': {
       disableWormholeElement: true,
     },
-    'ember-cli-terser': {
-      enabled: true,
```
Windvis, line 41: We really don't want this in dev builds, it slows down hot rebuilds a lot.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-      exclude: ['assets/vendor.js', 'flandersMunicipalities.ts'],
-    },
-    'ember-cli-htmlbars-inline-precompile': {
```
Windvis, line 44: This is superceded by ember-cli-htmlbars, we should also remove the dependency.

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
-const environment = process.env.EMBER_ENV;
-const IS_PROD = environment === 'production';
-const IS_TEST = environment === 'test';
-
 module.exports = function (defaults) {
   let app = new EmberApp(defaults, {
-    autoImport: {
-      exclude: ['flandersMunicipalities.ts'],
```
Windvis, line 12: This doesn't seem to be imported anywhere? Outdated config I guess?

ember-cli-build.js
```diff
@@ -2,48 +2,11 @@
...
-      webpack: {
-        node: {
-          global: true,
```
Windvis, line 15: The build didn't seem to fail with this removed, so I don't think we need it anymore.

ember-cli-build.js
aliokan: I guess it was use before using municipality from the server

aliokan (approved):
Approved! :pray:

PR #46, opened 2023-07-07
Ember data v4.10 update

Windvis:


app/controllers/map.ts
```diff
@@ -13,8 +13,8 @@ export default class MapComponent extends Controller {
   @service declare router: RouterService;
   declare model: ModelFrom<MapRoute>;
 
-  @tracked locationData: any = this.model.locationData.toArray();
-  @tracked agendaData: any = this.model.agendaData.toArray();
```
Windvis, line 17: This can never have worked properly. Controllers are only instanced once and there is no `this.model` at construction time. It's only available after the route executes the `setupController` method. This change is probably not correct either, but digging into that code is out of scope for this PR 😬.

app/controllers/agenda-items.ts
```diff
@@ -26,6 +29,10 @@ export default class AgendaItemsController extends Controller {
   @tracked loading = false;
   @tracked errorMsg = '';
 
+  setup() {
+    this.agendaItems = [...this.model.agendaItems];
```
Windvis, line 33: This can also be done in the `setupController` method of the route. I just did it like this because of personal preference, but I can change it around if needed.

app/controllers/map.ts
```diff
@@ -171,32 +171,17 @@ export default class MapComponent extends Controller {
...
-      .toArray()
-      .find((locationDataEntry: any) => {
-        return locationDataEntry.label === nameFromHover;
-      });
+    const munName = this.model.locationData.find((locationDataEntry: any) => {
+      return locationDataEntry.label === nameFromHover;
+    });
 
-    return munName.label
+    return munName?.label
```
Windvis, line 184: Not sure why I added this, either TS was complaining or I saw some exceptions at runtime.

app/controllers/agenda-items.ts
```diff
@@ -9,12 +9,15 @@ import { ModelFrom } from 'frontend-burgernabije-besluitendatabank/lib/type-util
...
 import MunicipalityListService from 'frontend-burgernabije-besluitendatabank/services/municipality-list';
+import AgendaItem from 'frontend-burgernabije-besluitendatabank/models/agenda-item';
 
 export default class AgendaItemsController extends Controller {
   @service declare router: RouterService;
   @service declare store: Store;
   @service declare keywordStore: KeywordStoreService;
   @service declare municipalityList: MunicipalityListService;
+
+  @tracked agendaItems: AgendaItem[] = [];
```
Windvis, line 20: Not sure if this is the correct type, but it acts like it so good enough for now I guess.

app/controllers/sessions/session.ts
```diff
@@ -2,11 +2,20 @@ import Controller from '@ember/controller';
...
 import SessionSessionRoute from '../../routes/sessions/session';
 import { sortObjectsByTitle } from 'frontend-burgernabije-besluitendatabank/utils/array-utils';
+import AgendaItem from 'frontend-burgernabije-besluitendatabank/models/agenda-item';
 
 export default class SessionsSessionController extends Controller {
   declare model: ModelFrom<SessionSessionRoute>;
 
   get agendaItemsSorted() {
-    return this.model.agendaItems.toArray().sort(sortObjectsByTitle);
+    let agendaItems = this.model.hasMany('agendaItems').value() as
```
Windvis, line 11: I switched to the hasMany reference API since it's not safe to access async relationships in getters in a sync way like the previous implementation. This version assumes the relationship is loaded already though.

app/controllers/agenda-items.ts
aliokan:
```suggestion
    this.agendaItems = [...this.model.agendaItems.toArray()];
```

Related to https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/46/files#diff-45b09f498e20729ac9a181d2ba5ae273a7466b2be5dfa51c4318155a93e5161eR26 changes. Here no complain because `agendaItems` is typed `any`, but would add a `toArray()` too.

aliokan (changes requested):
@Windvis Except the `Iterator` error, everything looks good.

package.json
```diff
@@ -94,7 +94,7 @@
     "ember-cli-terser": "^4.0.2",
     "ember-cli-typescript": "^5.2.1",
     "ember-concurrency": "^3.0.0",
-    "ember-data": "^3.28.13",
+    "ember-data": "^4.10.0",
```
Windvis, line 97: Should be `~4.10.0`

aliokan (approved):
Tested locally, everything looks good. LGTM! :rocket:

PR #47, opened 2023-07-07
Debounce search WIP

Windvis:
(on hold until it's clear how the search should work exactly)


Windvis:
Change of plans, we should remove the search on type instead and let the user click the search button.

PR #48, opened 2023-07-10
Implement Figma design

Dietr:
- [ ] Review CSS
- [ ] Implement Figma design
- [ ] Add legal pages
- [ ] Update legal pages content
- [ ] Accessibility review and fixes

### DIFF
--- a/app/routes/legal/accessibility-statement.js
+++ b/app/routes/legal/accessibility-statement.js
@@ -1,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class LegalAccessibilityStatementRoute extends Route {}
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -1,0 +1,22 @@
+// html {
+//   font-size: 50%;
+// }
+
+// ol {
+//   list-style-type: decimal;
+//   li {
+//     // display: inline;
+//     // float: left;
+//     margin: 0 12px;
+//     padding: 0 12px;
+//   }
+// }
+
+// path {
+//   stroke: white;
+//   stroke-width: 0.08rem;
+// }
+
+// cursor {
+//   cursor: pointer;
+// }
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -2,13 +2,22 @@
 
 <AuApp>
   <AppHeader />
-  <NavigationBar />
-
+  {{#if this.isIndex}}
+  <AuContentHeader
+    @titlePartOne="We geven je alle beslissingen"
+    @titlePartTwo="uit JOUW gemeente"
+  >
+    <img
+      sizes="50vw"
+      src={{"/assets/loket-header-1600.jpg"}}
+      alt="Foto van een laptop met daarop het vlaanderen logo."
+    />
+  </AuContentHeader>
+  {{/if}}
   <AuMainContainer as |m|>
-
     <m.content>
       {{outlet}}
-
     </m.content>
   </AuMainContainer>
+  <Footer />
 </AuApp>
--- a/app/templates/home.hbs
+++ b/app/templates/home.hbs
@@ -1,50 +1,46 @@
-<AuContentHeader
-  @titlePartOne="We geven je alle beslissingen"
-  @titlePartTwo="uit JOUW gemeente"
->
-  {{!-- <img
-    sizes="50vw"
-    src={{"/assets/loket-header-1600.jpg"}}
-    alt="Foto van een laptop met daarop het vlaanderen logo."
-  /> --}}
-</AuContentHeader>
-<AuBodyContainer class="au-u-flex au-u-flex--center au-u-flex--vertical-center">
 
-  <AuFormRow class="au-u-2-5">
-    <AuLabel
-      for={{"municipality"}}
-      class="au-u-h4 au-u-margin-bottom-small"
-      @inline={{true}}
-    >
-      In welke gemeente(s) wil je zoeken?
-    </AuLabel>
-    <section class="au-u-flex au-u-1-1"
-      {{did-insert this.resetLoading}}
+<AuBodyContainer @scroll={{true}}>
+  <section class="au-o-region-huge">
+    <div class="au-o-layout au-o-layout--small">
+      <AuLabel
+        for={{"municipality"}}
+        class="au-u-h5 au-u-margin-bottom-tiny"
+        @inline={{true}}
       >
-      <PowerSelect
-
-        @onChange={{this.handleMunicipalityChange}}
-        @selected={{this.selectedMunicipality}}
-        @options={{this.municipalities}}
-        @allowClear={{true}}
-        @searchField="label"
-        @placeholder={{"Alle gemeentes"}}
-        @noMatchesMessage={{"Geen gemeente gevonden"}}
-        @searchMessage="Aan het zoeken..."
-        @searchEnabled={{true}}
-        class="au-u-margin-right-small au-u-4-5"
-        as |singleSelect|
-      >
-        {{singleSelect.label}}
-      </PowerSelect>
-      <AuButton
-        class="au-u-1-5"
-        @loading={{this.loading}}
-        @width="block"
-        {{on "click" this.handleMunicipalitySelect}}
-      >Locatie(s) Opslaan</AuButton>
-    </section>
-  </AuFormRow>
+        In welke gemeente(s) wil je zoeken?
+      </AuLabel>
+      <div class="au-u-1-1 au-o-grid au-o-grid--tiny"
+        {{did-insert this.resetLoading}}
+        >
+        <div class="au-o-grid__item au-u-3-4@small">
+          <PowerSelect
+            @onChange={{this.handleMunicipalityChange}}
+            @selected={{this.selectedMunicipality}}
+            @options={{this.municipalities}}
+            @allowClear={{true}}
+            @searchField="label"
+            @placeholder={{"Alle gemeentes"}}
+            @noMatchesMessage={{"Geen gemeente gevonden"}}
+            @searchMessage="Aan het zoeken..."
+            @searchEnabled={{true}}
+            as |singleSelect|
+          >
+            {{singleSelect.label}}
+          </PowerSelect>
+        </div>
+        <div class="au-o-grid__item au-u-1-4@small">
+          <AuButton
+            @loading={{this.loading}}
+            @width="block"
+            {{on "click" this.handleMunicipalitySelect}}
+          >Locatie(s) Opslaan</AuButton>
+        </div>
+        <div class="au-o-grid__item au-u-text-center">
+          <AuHelpText @skin="secondary">Hu, gemeentes? <AuLink>Laat mij gewoon zoeken</AuLink></AuHelpText>
+        </div>
+      </div>
+    </div>
+  </section>
 </AuBodyContainer>
 
-<Footer />
+
--- a/app/templates/legal/accessibility-statement.hbs
+++ b/app/templates/legal/accessibility-statement.hbs
@@ -1,0 +1,50 @@
+{{page-title "Toegankelijkheidsverklaring"}}
+
+<AuContent class="au-u-3-4@medium">
+  <AuHeading>Toegankelijkheidsverklaring</AuHeading>
+  <p>Agentschap Binnenlands Bestuur streeft ernaar zijn websites en mobiele applicaties toegankelijk te maken, overeenkomstig het <a href="http://www.ejustice.just.fgov.be/eli/decreet/2018/12/07/2018032457/justel" target="_blank" rel="noopener noreferrer">Bestuursdecreet van 7 december 2018</a>.</p>
+  <p>Deze toegankelijkheidsverklaring is van toepassing op <a href="https://gelinkt-notuleren.vlaanderen.be">gelinkt-notuleren.vlaanderen.be</a></p>
+
+  <h2>Nalevingsstatus</h2>
+  <p>Deze web applicatie voldoet gedeeltelijk aan de <a href="https://www.w3.org/Translations/WCAG21-nl/" target="_blank" rel="noopener noreferrer">Richtlijnen voor Toegankelijkheid van Webcontent (WCAG) 2.1 Niveau AA</a>, omdat niet aan de onderstaande eis(en) is voldaan.</p>
+
+  <h2>Niet-toegankelijke inhoud</h2>
+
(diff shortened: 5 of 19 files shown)


app/styles/_shame.scss
Denperidge, line 1: While I genuinely love the name of this file, is there a reason for the merging of au-overrides and custom classes? While both are to a certain extent undesirable considering that we're using a framework, I do think there is value in having them be clearly outlined

app/templates/application.hbs
```diff
@@ -2,13 +2,22 @@
 
 <AuApp>
   <AppHeader />
-  <NavigationBar />
-
+  {{#if this.isIndex}}
```
Denperidge, line 5: Is there a more Ember-esque solution for this? Whether it be an outlet/yield/block situation or a variable that can easily be set in the route itself. The current structure  - whilst clean - does cause a bit of a unclarity

app/styles/components/_c-interface.scss
```diff
@@ -0,0 +1,104 @@
+.c-interface {
+  display: flex;
+  flex-wrap: wrap;
```
Denperidge, line 3: If possible, it might be worth changing some of the things in this file (like display: flex, flex-wrap wrap, and padding) to using au classes

app/styles/components/_c-infinite-list.scss
```diff
@@ -0,0 +1,7 @@
+.au-c-infinite-list {
+  > li + li {
+    margin-top: $au-unit;
```
Denperidge, line 3: Possibly replaceable using au- classes

app/styles/components/_c-description-list.scss
```diff
@@ -0,0 +1,22 @@
+.c-description-list {
+  --basis: #{$au-unit};
+  --gap: #{$au-unit-small};
+
+  display: flex;
```
Denperidge, line 5: Parts possibly replaceable using au classes

app/components/filter-sidebar.hbs
Denperidge, line 1: Is there a specific reason as to why the layout has been moved to the individual routes, seeing as they are meant to be the same? (If I understand correctly)

Denperidge (commented):
Aside from my comments (which can possibly be naught), it looks good!

app/components/infinite-list.hbs
```diff
@@ -1,15 +1,15 @@
 <div
-  class="{{@class}} overflow-y-scroll au-u-flex au-u-flex--column"
```
Windvis, line 2: I think this breaks the "auto load more items" functionality since the div doesn't scroll anymore. Not sure if it's simple to revert while keeping the styling improvements.

I think there are better ways to implement this than using a scroll event though so it might be worth simply redoing it in a seperate PR after merging this and consider it a known issue here?

app/components/vote-overview.ts
```diff
@@ -19,6 +20,47 @@ export default class VoteOverview extends Component<ArgsInterface> {
...
   }
 
+  get numberOfAbstentionsGraphValue() {
+    const totalValue =
+      (this.args.numberOfProponents || 0) +
+      (this.args.numberOfOpponents || 0) +
+      (this.args.numberOfAbstentions || 0);
+    const abstentionsValue = this.args.numberOfAbstentions || 0;
+
+    return (abstentionsValue / totalValue) * 158;
```
Windvis, line 30: What does the `158` mean exactly? Might be worth storing it in a descriptive const variable to make that more clear?

app/routes/legal/accessibility-statement.js
Windvis, line 1: I would remove these route files if they aren't used to fetch data. It shouldn't be needed and saves some kbs of bundle space 😄.

app/styles/components/_c-interface.scss
Dietr: Seems a bit complicated to me as the responsive behaviour is quite custom here. I guess it's possible but might cause conflicts with utility classes being the most specific.

app/styles/_shame.scss
Windvis: This is how we usually do it in other projects. `_shame.scss` is everything that ideally would be unneeded over time, either by upstreaming things or changing to different patterns. If it's for something very specific for the project that never will be replaced we usually move it to a different dedicated file.

So merging both into _shame makes sense to me 👍.

app/components/filter-sidebar.hbs
Dietr: Moved to be able to implement the mobile filter on the general overview.

tests/unit/routes/legal/accessibility-statement-test.js
Windvis, line 1: I would delete these too, I think that's the trend with "empty" test files at least.

app/styles/_shame.scss
Dietr: The _shame file is used for temporary overrides and fixes that can eventually move to the framework. In an ideal scenario this code will be refactored in the future. It's referring to https://appuniversum.github.io/ember-appuniversum/?path=/story/outline-architecture-and-namespaces--page

app/templates/home.hbs
```diff
@@ -1,50 +1,46 @@
...
+        </div>
+        <div class="au-o-grid__item au-u-1-4@small">
+          <AuButton
+            @loading={{this.loading}}
+            @width="block"
+            {{on "click" this.handleMunicipalitySelect}}
+          >Locatie(s) Opslaan</AuButton>
+        </div>
+        <div class="au-o-grid__item au-u-text-center">
+          <AuHelpText @skin="secondary">Hu, gemeentes? <AuButton @skin="link" {{on "click" this.handleMunicipalitySelect}}>Laat mij gewoon zoeken</AuButton></AuHelpText>
```
Windvis, line 39: Is this needed, I don't see it in the desktop version of the design?

app/styles/components/_c-infinite-list.scss
Dietr: I prefer defining it here as the AU list doesn't cover the design spec. Will see if I can extend the list component in the future.

app/styles/components/_c-description-list.scss
Dietr: See other comment about au utility classes.

app/templates/application.hbs
```diff
@@ -2,13 +2,22 @@
...
+  <AuContentHeader
+    @titlePartOne="We geven je alle beslissingen"
+    @titlePartTwo="uit JOUW gemeente"
+  >
+    <img
+      sizes="50vw"
+      src={{"/assets/loket-header-1600.jpg"}}
```
Windvis, line 12:
```suggestion
      src="/assets/loket-header-1600.jpg"
```
The srcsets might also be nice? (but can also be done in separate smaller PRs).

app/templates/legal/accessibility-statement.hbs
```diff
@@ -0,0 +1,27 @@
+{{page-title "Toegankelijkheidsverklaring"}}
+
+<AuContent class="au-u-3-4@medium">
+  <AuHeading>Toegankelijkheidsverklaring</AuHeading>
+  <p>Agentschap Binnenlands Bestuur streeft ernaar zijn websites en mobiele applicaties toegankelijk te maken, overeenkomstig het <a href="http://www.ejustice.just.fgov.be/eli/decreet/2018/12/07/2018032457/justel" target="_blank" rel="noopener noreferrer">Bestuursdecreet van 7 december 2018</a>.</p>
```
Windvis, line 5: I think we can use AuLinkExternal in these files?

```suggestion
  <p>Agentschap Binnenlands Bestuur streeft ernaar zijn websites en mobiele applicaties toegankelijk te maken, overeenkomstig het <AuLinkExternal href="http://www.ejustice.just.fgov.be/eli/decreet/2018/12/07/2018032457/justel">Bestuursdecreet van 7 december 2018</AuLinkExternal>.</p>
```

app/templates/legal/accessibility-statement.hbs
```diff
@@ -0,0 +1,27 @@
...
+
+  <h2>Nalevingsstatus</h2>
+  <p>Deze web applicatie voldoet aan de <a href="https://www.w3.org/Translations/WCAG21-nl/" target="_blank" rel="noopener noreferrer">Richtlijnen voor Toegankelijkheid van Webcontent (WCAG) 2.1 Niveau AA</a>.</p>
+
+  <h2>Opstelling van deze toegankelijkheidsverklaring </h2>
+  <p>Deze toegankelijkheidsverklaring is opgesteld op 11/07/2023.</p>
+  <p>Voor deze toegankelijkheidsverklaring werd de toegankelijkheid van de vermelde web applicatie beoordeeld door het Agentschap Binnenlands Bestuur.</p>
+
+  <h2>Feedback en contactgegevens</h2>
+  <p>Voor alle vragen en opmerkingen over de toegankelijkheid van de vermelde websites van het Agentschap Binnenlands Bestuur kunt u contact opnemen via <a href="mailto:binnenland@vlaanderen.be">binnenland@vlaanderen.be</a></p>
```
Windvis, line 16: `AuLinkExternal` should work for the mail as well I think, but otherwise maybe the `au-c-link` class might improve the styles a bit?

Windvis (commented):
Looks a loooot better already, thanks!

Some initial feedback.

app/components/infinite-list.hbs
Dietr: Missed the auto scroll. Should be fixed now.

app/components/vote-overview.ts
```diff
@@ -19,6 +20,47 @@ export default class VoteOverview extends Component<ArgsInterface> {
...
+  @action
+  bake() {
+    [
+      { key: '--value-abstentions', val: this.numberOfAbstentionsGraphValue },
+      { key: '--value-opponents', val: this.numberOfOpponentsGraphValue },
+      { key: '--value-proponents', val: this.numberOfProponentsGraphValue },
+    ].forEach(({ key, val }) => {
+      document.documentElement.style.setProperty(key, val.toString());
+    });
+  }
```
Windvis, line 62: I think this can also be called from the constructor to have the same effect without having to use `{{did-insert}}`.

app/components/vote-overview.hbs
```diff
@@ -1,6 +1,12 @@
-<C3Chart
-  @data={{this.data}}
-  @color={{this.color}}
-  @pie={{this.pie}}
-  @legend={{this.legend}}
-/>
+<div class="c-chart">
```
Windvis, line 1: Looks good 👍 The only thing missing seems to be the hover effect that the previous chart did. I don't really think it's needed, but Jodi might need to decide if it's acceptable like this?

app/components/vote-overview.ts
Dietr: Bit of a magic number. The length of the stroke-dasharray in a circle is 158. Will put it in a variable.

app/templates/application.hbs
Dietr: Do you have an example of how this works? Or do I simply create two components that wrap the page content and set those on each template?

Windvis: Yea, that's indeed what I meant. But I don't see it as a blocker here.

app/templates/detail.hbs
```diff
@@ -1,243 +1,250 @@
...
+              {{#if this.hasVotes}}
+                <section class="c-accordion-holder">
+                  <AuBadge @icon="message" @skin="brand" />
+                  <AuAccordion
+                    @reverse={{true}}
+                    @iconOpen={{"nav-down"}}
+                    @iconClosed={{"nav-right"}}
+                    @buttonLabel={{"Stemming"}}
+                  >
+                    {{#each this.model.agendaItem.handledBy.hasVotes as |vote|}}
```
Windvis, line 123: The design shows a "Openbare stemming" text above the chart.

app/templates/detail.hbs
```diff
@@ -1,243 +1,250 @@
...
+
+                    <div class="c-accordion-holder c-accordion-holder--top">
+                      <AuAccordion
+                        @reverse={{true}}
+                        @iconOpen={{"nav-down"}}
+                        @iconClosed={{"nav-right"}}
+                        @buttonLabel={{"Wie stemde?"}}
+                        @subtitle="Meer informatie over stemmers en hun partijen"
+                      >
+                        <AuTable>
```
Windvis, line 139: Not sure if all mobile fixes are planned for this PR, but it splits into multiple tables on small devices.

app/templates/application.hbs
```diff
@@ -2,13 +2,22 @@
...
+  </AuContentHeader>
+  {{/if}}
   <AuMainContainer as |m|>
-
     <m.content>
       {{outlet}}
-
     </m.content>
   </AuMainContainer>
+  <Footer />
```
Windvis, line 22: The footer seems to be sticky which takes up quite some space on smaller screens. I guess it makes sense on the infinite scroll page since it otherwise would keep being pushed down until the end of the list, but on the details page it might be nice if it would just scroll with the content maybe?

app/templates/application.hbs
```diff
@@ -2,13 +2,22 @@
...
-
+  {{#if this.isIndex}}
+  <AuContentHeader
+    @titlePartOne="We geven je alle beslissingen"
+    @titlePartTwo="uit JOUW gemeente"
+  >
+    <img
+      sizes="50vw"
+      src={{"/assets/loket-header-1600.jpg"}}
+      alt="Foto van een laptop met daarop het vlaanderen logo."
```
Windvis, line 13:
```suggestion
      alt=""
      {{!template-lint-disable require-valid-alt-text}}
```

app/templates/detail.hbs
```diff
@@ -1,243 +1,250 @@
...
-        >
-          <AuIcon class="au-u-margin-right-small" @size="large" @icon="check" />
-          <p>{{this.model.agendaItem.session.dateFormatted}}</p>
-        </section>
-        <section
-          class="au-u-flex au-u-muted au-u-margin-top-tiny au-u-u-flex--vertical-center"
+        </div>
+        <article
+          id="detail-left-section"
+          class="au-o-grid__item au-u-4-6@small"
```
Windvis, line 31: This might a bit to small on tablet portrait mode? Maybe `@medium` would be better?

app/templates/application.hbs
Dietr: You're right but the linter won't allow a empty alt tag so can't commit this change :-)

Windvis: Yea, but the disable comment in the snippet should fix that?

Windvis (approved):
Thanks!

PR #49, opened 2023-07-10
Enable embroider

Windvis:
I wanted to give Embroider a try here since it has the potential of reducing the bundle size compared to classic builds and everything seems to work fine on the strictest settings 👍.

Embroider is the new build pipeline that uses (at the moment) Webpack. It uses static code analysis to remove unneeded code from the app which is especially nice for public facing websites.

More info: https://github.com/embroider-build/embroider


ember-cli-build.js
```diff
@@ -9,5 +9,14 @@ module.exports = function (defaults) {
...
-  return app.toTree();
+  const { Webpack } = require('@embroider/webpack');
+
+  return require('@embroider/compat').compatBuild(app, Webpack, {
+    staticAddonTestSupportTrees: true,
+    staticAddonTrees: true,
+    staticHelpers: true,
+    staticModifiers: true,
+    staticComponents: true,
+    splitAtRoutes: ['map'],
```
Windvis, line 20: This splits off _some_ of the depenendencies that are only used in the map route. They will be loaded when the user enters the route so the main bundle size is reduced.

aliokan: Thank you! :pray:

aliokan (approved):
LGTM! :rocket:

aliokan:
## Before

image

## After

image

PR #50, opened 2023-07-11
Added pre-commit, moved dependencies to devDependencies, npm script docs

Denperidge:
- There were dependencies set under regular dependencies, but they were devDependencies: the end user does not need them installed
- Fixed git url in package.json
- npm run lint is now run before committing
- Documented npm scripts
- Removed Ember-Data v3 documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,21 @@ cd frontend-burgernabije-besluitendatabank
 npm install
 ```
 
-From there, you can use `npm run dev` in order to use the mock API, `npm run prod` in order to use a local BNB api, or `npm run --proxy=https://burgernabije-besluitendatabank-dev.s.redhost.be/` to run using the BNB API on [the external dev server](https://burgernabije-besluitendatabank-dev.s.redhost.be/).
+From there, you can use `npm run dev:proxy`, or any of the following npm scripts:
+
+
+|  npm run...  |  Description |
+| ------------ | ------------ |
+| build        | Creates a production-ready static build |
+| lint         | Runs the linter and returns any errors/warnings. *Is run automatically before committing, cancelling the commit on error* |
+| lint:fix     | Runs the linter, attempting to automatically fix any errors |
+| dev          | Run a development server with the mock api |
+| dev:proxy    | Run a development server with a proxy to [the external dev server](https://burgernabije-besluitendatabank-dev.s.redhost.be/) |
+| start        | Run a development server with the mock api |
+| prod         | Run a production server |
+| test         | Run the linter and then ember tests |
+
+
 
 ### Automated builds
 
@@ -74,6 +88,3 @@ The levels are as follows:
 
 ![Data Alignment Analysis]("./../docs/analysis-bnb.png)
 
-### Ember-Data@3
-
-For some reason beyond human comprehension, Ember-Data refused to work as is documented in the _latest documentation_ when using v4, but it did on v3. So guess what. It's v3 now.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -19,7 +19,21 @@ cd frontend-burgernabije-besluitendatabank
...
+|  npm run...  |  Description |
+| ------------ | ------------ |
+| build        | Creates a production-ready static build |
+| lint         | Runs the linter and returns any errors/warnings. *Is run automatically before committing, cancelling the commit on error* |
+| lint:fix     | Runs the linter, attempting to automatically fix any errors |
+| dev          | Run a development server with the mock api |
+| dev:proxy    | Run a development server with a proxy to [the external dev server](https://burgernabije-besluitendatabank-dev.s.redhost.be/) |
+| start        | Run a development server with the mock api |
+| prod         | Run a production server |
+| test         | Run the linter and then ember tests |
```
Windvis, line 34: I'm not sure about this. Just looking at the scripts config in the package.json file gives you the almost same info and it's always up to date.

package.json
```diff
@@ -128,12 +135,7 @@
...
-  "dependencies": {
-    "c3": "0.7.20",
-    "d3": "6.6.0",
-    "d3-geo": "^3.1.0",
-    "date-fns": "^2.29.3",
-    "postscribe": "^2.0.8",
-    "topojson": "^3.0.2"
-  }
+  "pre-commit": [
+    "lint"
```
Windvis, line 139: How does this work exactly? How does git know it should run this before committing?

Also, while I think it's fine, I feel this might be controversial so not sure if we can roll this out before it's discussed 😬 (I don't make the rules, sorry).

README.md
```diff
@@ -74,6 +88,3 @@ The levels are as follows:
 
 ![Data Alignment Analysis]("./../docs/analysis-bnb.png)
 
-### Ember-Data@3
-
-For some reason beyond human comprehension, Ember-Data refused to work as is documented in the _latest documentation_ when using v4, but it did on v3. So guess what. It's v3 now.
```
Windvis, line 79: Thanks, I forgot about that, sorry 😄.

Windvis (commented):
Thanks!

README.md
Denperidge: That is fair! I've mainly added it because it might help onboarding, especially since some of the info is a bit more obscured (for example that dev uses a mock API and dev:proxy's url is our external dev server). It also obscures npm run lint:hbs:fix and the other sub-scripts. However, if the maintenance is too big of a possible maintenance hazard, I can understand if it can't make it in!

package.json
Denperidge: It uses an npm package! https://www.npmjs.com/package/pre-commit


That is fair! I can turn this into a draft for the time being?

Windvis: > It uses an npm package! https://www.npmjs.com/package/pre-commit

Oh, nice, I didn't see it when I glanced at the other changes, that makes sense 👍.

> That is fair! I can turn this into a draft for the time being?

I'm not sure. If everyone on the team is on board, feel free to keep it, but be warned that there might be complaints in the future 😄. I wouldn't let it block the other changes though. So we either just go with it and roll back if needed, or split it to a separate PR.

Denperidge: I guess it's a "do you prefer to know immediately or do you prefer to know when the CI fails" 😭

Windvis: Let's just do it. We'll deal with the fallout when it happens 😄.

Windvis (approved):

PR #51, opened 2023-07-11
TypeScript warning fixes & subsequent refactoring

Denperidge:
Important refactors:
*Routes - session*
I've simplified the code a bit, because the previous code seems like a roundabout cache. Caching should be handled by Ember-Data

*Controllers - Municipality*
Even with municipality's page being in limbo,
the code in the controller was unused


config/environment.d.ts
Windvis, line 1: Is this file needed? I think we only need the `app/config/environment.d.ts` version?

app/config/environment.d.ts
```diff
@@ -3,8 +3,8 @@
  *    import config from 'my-app/config/environment'
  */
 declare const config: {
-  VO_HEADER_WIDGET_URL: any;
-  VO_FOOTER_WIDGET_URL: any;
+  VO_HEADER_WIDGET_URL: string;
+  VO_FOOTER_WIDGET_URL: string;
```
Windvis, line 7: Are these still needed? I think we're just using the Appuniversum based header and footer?

app/controllers/map.ts
```diff
@@ -6,24 +6,36 @@ import { tracked } from '@glimmer/tracking';
...
+    type: 'FeatureCollection',
+    features: [],
+  };
+  @tracked agendaData: Array<AgendaItemModel> = [];
+
+  @tracked provincesData?: FeatureCollection = {
+    type: 'FeatureCollection',
+    features: [],
+  };
+  //@tracked provincesData: Feature[] = [];
```
Windvis, line 30: I would delete things if they aren't used.

types/ember-data/types/registries/model.d.ts
```diff
@@ -2,5 +2,7 @@
  * Catch-all for ember-data.
  */
 export default interface ModelRegistry {
+  // Allow any for the sake of catch-all
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [key: string]: any;
```
Windvis, line 7: It seems most of our models are missing the ModelRegistry code: https://docs.ember-cli-typescript.com/ts/using-ts-effectively#ember-data-lookups I wonder if this catch-all is needed if we add those. Or maybe it can be set to the base `Model` class?

README.md
```diff
@@ -19,21 +19,7 @@ cd frontend-burgernabije-besluitendatabank
 npm install
 ```
 
-From there, you can use `npm run dev:proxy`, or any of the following npm scripts:
-
-
-|  npm run...  |  Description |
```
Windvis, line 25: Wait, was this intentional? Merge conflict whoopsie?

app/controllers/map.ts
```diff
@@ -6,24 +6,36 @@ import { tracked } from '@glimmer/tracking';
 import * as d3 from 'd3';
 import { geoMercator, geoPath } from 'd3-geo';
 import MapRoute from 'frontend-burgernabije-besluitendatabank/routes/map';
-import * as topojson from 'topojson';
+import { Feature, FeatureCollection } from 'geojson';
```
Windvis, line 9: No blocker for this PR, but I know that "type-only" imports are a thing in TS and I think it might make more clear that something is only imported for typing reasons. https://typescript-eslint.io/rules/consistent-type-imports/

Maybe this is something we could enable in the future if everyone agrees?

Windvis (commented):
Nice work! Some nitpicks / questions.

app/routes/detail.ts
```diff
@@ -4,6 +4,16 @@ import { service } from '@ember/service';
 import KeywordStoreService from 'frontend-burgernabije-besluitendatabank/services/keyword-store';
 import { sortObjectsByTitle } from 'frontend-burgernabije-besluitendatabank/utils/array-utils';
 
+interface detailParams {
```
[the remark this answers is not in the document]
Denperidge: You are correct!

app/routes/detail.ts
```diff
@@ -4,6 +4,16 @@ import { service } from '@ember/service';
...
+  id: string;
+}
+
+interface formattedTableVoteInterface {
```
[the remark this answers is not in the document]
Denperidge: Changed!

app/controllers/map.ts
Denperidge: Done!

app/config/environment.d.ts
Denperidge: I think you're correct!

Windvis:
It seems we're missing TS blueprints (which normally set up the needed registries). Historically those came from the [ember-cli-typescript-blueprints addon](https://github.com/typed-ember/ember-cli-typescript-blueprints/blob/8d1a938644a41f504bdef0242248367e1c827d12/blueprints/model/files/__root__/__path__/__name__.ts#L10-L14) but Ember 4.12 ships them as well. So we should update. I'll do that in a separate PR.

types/ember-data/types/registries/model.d.ts
Denperidge: I've tried Model, but it gave a bunch of other errors. I'll try to implement the ModelRegistery!

README.md
Denperidge: I am my own biggest enemy :(

app/controllers/map.ts
Denperidge: Will make a message in rocketchat about this!

types/ember-data/types/registries/model.d.ts
Windvis: Yea, I don't think Model will work since everything is a subclass and adds extra fields. But I don't like the `any` since that is how it went unnoticed that the model registry entries were missing 😄.

So I would indeed add the registries (doesn't have to be this PR) and then we can maybe change this to `Model` (or `unknown`?) so that code fails if the registry entry is missing.

config/environment.d.ts
Denperidge: Correct and removed!

types/ember-data/types/registries/model.d.ts
Denperidge: I'll indeeed move it it to its own Issue so this PR can be finished!

README.md
```diff
@@ -88,3 +88,69 @@ The levels are as follows:
 
 ![Data Alignment Analysis]("./../docs/analysis-bnb.png)
 
+
+### Filtering
```
Windvis, line 92: This comes from the other PR I think? I would remove it here since it's not really related to the other changes.

Denperidge: Oh lord I like how I did the pull request equivalent of dropping glass on the floor

Windvis (approved):

PR #52, opened 2023-07-11
Feature/detail enhance municipality button behavior

aliokan:
This PR contains: 
- model type fixes
- Update relationship between `agenda-item`, `session` and `governing-body` (Related to the **BREAKING CHANGE** API lblod/app-burgernabije-besluitendatabank#17)
- Use `administrative-unit.location.label` everywhere to resolve municipality name instead of `administrative-unit.name`
- Remove the not working filtering `:has:` from query
- Redirect users to the agendaItems list, filtered specifically for the selected municipality when clicking on municipality Button


app/models/agenda-item.ts
```diff
@@ -8,9 +14,15 @@ export default class AgendaItemModel extends Model {
...
-  @belongsTo('session', { async: true, inverse: 'agendaItems' })
-  declare session?: SessionModel;
+  @hasMany('session', { async: true, inverse: 'agendaItems' })
+  declare sessions?: AsyncHasMany<SessionModel>;
 
   @belongsTo('agenda-item-handling', { async: true, inverse: null })
-  declare handledBy?: AgendaItemHandlingModel;
+  declare handledBy?: AsyncBelongsTo<AgendaItemHandlingModel>;
+
+  get session() {
```
Denperidge, line 23: **Question:** This getter is meant to then return the first session it can find for the agenda-item, right?

app/models/governing-body.ts
```diff
@@ -11,6 +16,18 @@ export default class GoverningBodyModel extends Model {
...
-  @belongsTo('session', { async: true, inverse: 'governingBody' })
-  declare session: AsyncBelongsTo<SessionModel>;
+  @hasMany('session', { async: true, inverse: 'governingBody' })
+  declare sessions: AsyncHasMany<SessionModel>;
+
+  @belongsTo('governing-body', {
+    async: false,
+    inverse: 'hasTimeSpecializations',
+  })
+  declare isTimeSpecializationOf: GoverningBodyModel;
```
Denperidge, line 26: **Question:** Can you clarify what timeSpecialization is for/refers to?

app/models/session.ts
```diff
@@ -14,32 +14,28 @@ export default class SessionModel extends Model {
...
 
-  @hasMany('agenda-item', { async: true, inverse: 'session' })
+  @hasMany('agenda-item', { async: true, inverse: 'sessions' })
   declare agendaItems: AsyncHasMany<AgendaItemModel>;
 
-  @belongsTo('governing-body', { async: false, inverse: 'session' })
+  @belongsTo('governing-body', { async: false, inverse: 'sessions' })
   declare governingBody: GoverningBodyModel;
 
   get name() {
```
Denperidge, line 23: **Suggestion:** Should we perhaps change this getters name? It's a bit ambiguously named

Denperidge (approved):
A few things are unclear to my tired brain, but nothing that would or should stop this PR!

app/models/governing-body.ts
aliokan: There are 2 types of governing bodies:
- the abstraction (undated)
- the time specified

This is described in the mandatendatabank specification: https://data.vlaanderen.be/doc/applicatieprofiel/mandatendatabank/html/overview.jpg

From a Session, we are supposed to go to `governingBody.isTimeSpecializationOf.administrativeUnit`. 

I discover it last Friday (https://chat.semte.ch/group/X8kirHFphoFXfBKXZ?msg=jxCt2wxns58ssKpuP)

Denperidge (approved):
New changes, second approve!

app/models/agenda-item.ts
```diff
@@ -8,9 +14,19 @@ export default class AgendaItemModel extends Model {
...
   @belongsTo('agenda-item-handling', { async: true, inverse: null })
-  declare handledBy?: AgendaItemHandlingModel;
+  declare handledBy?: AsyncBelongsTo<AgendaItemHandlingModel>;
+
+  /**
+   * @returns the first session with hasMunicipality == true
+   * This is the session we want to use to resolve municipality & governingBody name
+   */
+  get session() {
+    return this.sessions?.slice().find((session) => {
```
Windvis, line 28: Accessing async relationships in a sync way like this is deprecated. `.slice()` also triggers deprecations.

If the relationship needs to stay async we can use the [reference APIs](https://api.emberjs.com/ember-data/5.1/classes/Model/methods/hasMany?anchor=hasMany), assuming the relationship was already loaded (via includes for example.

```suggestion
    return this.hasMany('sessions').value()?.slice().find((session) => {
```

TS is complaining though, and I'm not sure what the issue is, I think it's another bug in the DT types.

app/routes/detail.ts
```diff
@@ -110,11 +101,16 @@ export default class DetailRoute extends Route {
...
       filter: {
-        session: {
+        sessions: {
           'governing-body': {
-            'administrative-unit': {
-              name: agendaItem.session?.get('municipality') || undefined,
+            'is-time-specialization-of': {
+              'administrative-unit': {
+                location: {
+                  label: agendaItem.session?.get('municipality') || undefined,
```
Windvis, line 111: `.get` shouldn't be needed if the session getter returns the record instead of the proxy.

app/controllers/map.ts
```diff
@@ -78,8 +78,7 @@ export default class MapComponent extends Controller {
         this.agendaData.forEach((agendaItem: AgendaItemModel) => {
           const agendaItemLocation = agendaItem
             .get('session')
-            ?.get('governingBody')
-            ?.get('administrativeUnit')?.name;
+            ?.get('municipality');
```
Windvis, line 81: `.get` is usually a code smell and means that the async relationship isn't handled correctly. I think it might not be needed here anymore if the session getter is updated.

app/models/agenda-item.ts
Windvis: Did some digging and it seems to be this issue: https://github.com/typed-ember/ember-cli-typescript/issues/1416

The mentioned workaround in the last comment seems to work.
```suggestion
    return (this as AgendaItemModel).hasMany('sessions')?.find((session) => {
```

aliokan: Type `this`, TS is magic :heart_eyes:

aliokan:
@Windvis ready for the next round! :ping_pong:

Windvis (approved):
👍

PR #55, opened 2023-07-13
Update the favicon

Windvis:
This change the favicon to an svg so that the white background is removed.

Before/after (in Firefox):
image


madnificent (approved):
Side chat with @Windvis indicitas this is the same way as vlaanderen.be shows the favicon.  It looks a bit better and it's in line with broader branding.

aliokan:
LGTM! 🚀

aliokan (approved):

PR #56, opened 2023-07-13
Fix(AgendaItems) InfiniteList data loading

aliokan:
- use `agendaItems` from `controller` instead of `model`


Windvis (approved):
Good catch, probably a merge conflict gone wrong.

PR #57, opened 2023-07-13
InfiniteList hide load more & AgendaItems result #

Denperidge:
image
Woah

Note: not fully testable until the LoadMore issue has been fixed

### DIFF
--- a/app/templates/agenda-items.hbs
+++ b/app/templates/agenda-items.hbs
@@ -45,9 +45,15 @@
                 </AuAlert>
               </div>
             {{else}}
+              {{#if @model.count}}
+                <p class="au-u-margin-left au-u-margin-top-large au-u-margin-bottom-small au-u-h4 au-u-medium">{{@model.count}} Zoekresultaten</p>
+              {{/if}}
+
               <InfiniteList
                 @loadMore={{this.loadMore}}
                 @isLoading={{this.isLoadingMore}}
+                @itemsShown={{this.agendaItems.length}}
+                @itemsAmount={{@model.count}}
               >
                 {{#each @model.agendaItems as |item|}}
                 <li>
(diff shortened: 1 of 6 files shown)


app/templates/agenda-items.hbs
```diff
@@ -45,9 +45,15 @@
                 </AuAlert>
               </div>
             {{else}}
+              {{#if @model.count}}
+                <p class="au-u-margin-left au-u-margin-top-large au-u-margin-bottom-small au-u-h4 au-u-medium">{{@model.count}} Zoekresultaten</p>
```
Windvis: I would remove the bottom margin since the text and the items are a bit closer together in the design.

```suggestion
                <p class="au-u-margin-left au-u-margin-top-large au-u-h4 au-u-medium">{{@model.count}} Zoekresultaten</p>
```

Windvis: A border is also shown which isn't there in the design. It comes from here: https://github.com/lblod/frontend-burgernabije-besluitendatabank/blob/93e5cd623536bc42e8cb436464e567f8f8a9f6ac/app/styles/_shame.scss#L207-L209

I would remove it, I doesn't seem to be used anywhere else at the moment (@Dietr?)

Windvis (commented):
Nice. Some small remarks but other than that, good to go 👍.

app/templates/agenda-items.hbs
Denperidge: Both implemented! I moved the border-top definition to its own class for now, in case Dieter still planned on using it

Windvis (approved):

app/templates/agenda-items.hbs
Dietr: Yes, border was there to adapt to the scrolling which is not shown in the design. Without the border the text gets cut of with a white background which looks a bit messy.

Dietr: No worries, will add it back when I do some finetuning of the design. I don't have figma editing permissions but will ask them as it would be easier to align on the design.

PR #58, opened 2023-07-13
Disable the "search-on-type" functionality

Windvis:
This removes the "search-on-type" functionality and replaces it with a submit button. Users can now click the button (or press enter) to trigger the search.

### DIFF
--- a/app/components/filters/filter.ts
+++ b/app/components/filters/filter.ts
@@ -3,17 +3,19 @@ import RouterService from '@ember/routing/router-service';
 import { service } from '@ember/service';
 import { get } from '@ember/object';
 
-interface ArgsInterface {
-  id: string;
-  queryParam: string;
-  queryParamsA: string;
-  queryParamsB: string;
+interface Signature {
+  Args: {
+    id: string;
+    queryParam: string;
+    queryParamsA: string;
+    queryParamsB: string;
 
-  searchField: string;
-  options: Array<object>;
+    searchField: string;
+    options: Array<object>;
+  };
 }
 
-export default class FilterComponent extends Component<ArgsInterface> {
+export default class FilterComponent<S = Signature> extends Component<S> {
   @service declare router: RouterService;
 
   /**
--- a/app/components/filters/text-filter.hbs
+++ b/app/components/filters/text-filter.hbs
@@ -1,29 +1,28 @@
-<Filters::Filter @id={{@id}} @label={{@label}} @labelHidden={{@labelHidden}} @rowClass={{@rowClass}}>
-    <AuInput
+<form {{on "submit" this.handleSubmit}}>
+  <Filters::Filter
+    @id={{@id}}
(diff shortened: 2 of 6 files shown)


app/components/filters/filter.ts
```diff
@@ -3,17 +3,19 @@ import RouterService from '@ember/routing/router-service';
 import { service } from '@ember/service';
 import { get } from '@ember/object';
 
-interface ArgsInterface {
-  id: string;
-  queryParam: string;
-  queryParamsA: string;
-  queryParamsB: string;
+interface Signature {
```
Windvis, line 6: I updated this to match the setup in the ember-cli-ts documentation: https://docs.ember-cli-typescript.com/ember/components

app/components/filters/filter.ts
```diff
@@ -3,17 +3,19 @@ import RouterService from '@ember/routing/router-service';
...
 
-  searchField: string;
-  options: Array<object>;
+    searchField: string;
+    options: Array<object>;
+  };
 }
 
-export default class FilterComponent extends Component<ArgsInterface> {
+export default class FilterComponent<S = Signature> extends Component<S> {
```
Windvis, line 18: I have no idea what I'm doing, but this seems to use the local `Signature` as a backup so that the other filter components don't fail typechecking.

app/components/filters/text-filter.hbs
```diff
@@ -1,29 +1,28 @@
-<Filters::Filter @id={{@id}} @label={{@label}} @labelHidden={{@labelHidden}} @rowClass={{@rowClass}}>
-    <AuInput
+<form {{on "submit" this.handleSubmit}}>
```
Windvis, line 1: Using form + submit event allows users to press "enter" to start searching.

app/components/filters/text-filter.ts
```diff
@@ -1,14 +1,29 @@
...
+export interface Signature {
+  Args: {
+    id: string;
+    queryParam: string;
+    value: string;
+  };
+}
+
+export default class TextFilterComponent extends FilterComponent<Signature> {
+  @localCopy<string>('args.value', '') declare value: string;
```
Windvis, line 14: This allows us to use a local 'value' property that still updates when the argument changes. We don't want to update the argument directly as that would update the QP and trigger the model hook.

Alternatively we could still update the QP value directly but then we would have to remove the `refreshModel` value and trigger the update with `this.router.refresh('route-name')`. The downside of that is that if a user types in the field  and doesn't submit, and refreshes the page, the value would still be there which seemed weird since they didn't press search yet. So I opted for the localCopy version.

app/components/filters/filter.ts
Windvis: I meant subclassing our own components to share functionality that way.

aliokan (approved):
Look good, LGTM! :rocket:

PR #60, opened 2023-07-14
Various small design tweaks

Dietr:
- improved spacing on infinite list and decision cards: not following the design exactly but there's ample white space available and we should allow as much content as possible here
- add infinite list divider: the overflow scrolling needs some kind of divider, otherwise the scrolling cuts of against a white background
- fixed a bug where the keyword was showing an empty space when not available
- hide help link in the app header: not available right now so hidden until there's a valid link

### DIFF
--- a/app/components/app-header.hbs
+++ b/app/components/app-header.hbs
@@ -2,15 +2,5 @@
   @brandLink={{"https://www.vlaanderen.be/nl"}}
   @homeRoute="home"
   @appTitle="Bobcat"
->
-  {{!-- TODO: provide a valid help link --}}
-  {{!-- <li class="au-c-list-horizontal__item">
-    <AuLinkExternal
-      href="#!"
-      @skin="secondary"
-      @icon="question-circle"
-    >
-      Help
-    </AuLinkExternal>
-  </li> --}}
-</AuMainHeader>
+/>
+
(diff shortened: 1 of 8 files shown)


app/components/app-header.hbs
```diff
@@ -3,13 +3,14 @@
   @homeRoute="home"
   @appTitle="Bobcat"
 >
-  <li class="au-c-list-horizontal__item">
+  {{!-- TODO: provide a valid help link --}}
```
Windvis, line 6: Will there even be a help link? 😄 Seems like a straightforward app. Unless it's a "rule" that every app needs one?

I would just delete the code. We can add it again if needed.

app/styles/_shame.scss
```diff
@@ -207,4 +207,11 @@ cursor {
 .u-overflow-y-scroll {
   overflow-y: scroll;
   height: 100%; 
+}
+
+.u-overflow-divider {
+  * + & {
+    border-top: .1rem solid var(--au-gray-200);
+    margin-top: $au-unit;
```
Windvis, line 215: nitpick, but on tablet landscape the scroll area is still fairly small. 
image

I think the top part can be condensed a bit more maybe?

Dietr: True, reduced it a bit more. Still not ideal as font-sizes are also a bit big here. Will explore further in a separate PR.

Windvis: Awesome, thanks!

Windvis (approved):

PR #61, opened 2023-07-14
Cleanup 🗑️

Windvis:
A bunch of small changes to remove unused or unneeded things.


config/environment.js
```diff
@@ -28,12 +16,7 @@ module.exports = function (environment) {
     },
   };
 
-  if (environment === 'production') {
-    ENV.yappConfigKey = 'dev';
```
Windvis, line 32: No idea what this is / was.

config/environment.js
```diff
@@ -28,12 +16,7 @@ module.exports = function (environment) {
...
-    ENV.API_URL = 'https://burgernabije-besluitendatabank-dev.s.redhost.be/api';
```
Windvis, line 33: API_URL wasn't used anywhere anymore.

app/components/vote-overview.ts
```diff
@@ -1,6 +1,4 @@
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
-import * as d3 from 'd3';
```
Windvis, line 3: This one has the most impact on the app. Now only the map route imports d3 so Embroider can move it into the map route bundle which means the main bundle is a little bit smaller again 🎉.

aliokan (approved):
Great! :pray:

PR #66, opened 2023-07-18
Moved vote code to vote-overview

Denperidge:
Currently, every vote (both locally and on redhost) is 0. So until this can be properly tested, I'm letting it be a draft PR. But theoretically it should work!


Denperidge:
Tested using /agendapunt/b356b180-1780-11ee-abd8-f97e130a97cf !

PR #69, opened 2023-07-20
Removed map route & uninstalled its packages

Denperidge:
I've tagged the last commit that contained the map route, so that if/once the map gets re-added, the code is easy to find!


LBONTEN (approved):
That is a lot of lost work , for the time being at least

PR #70, opened 2023-07-20
Fixed InfiniteList attempting to load more after scrolling down after reaching the bottom

Denperidge:


LBONTEN (approved):
Looks good to me

PR #71, opened 2023-07-20
Fixed extra sessions not loading on sessions.index

Denperidge:
I didn't set out to fix this but it happened


LBONTEN (approved):
It's a kind of magic

PR #74, opened 2023-07-21
Refactored agenda-items route & controller

Denperidge:
Simplifies the agenda-items code, removes the now unneeded caching thing, and fixes the empty queryParam thing! Yay

Removed:
    - (Unused) sort queryParam
    - (Unused) AgendaItemsRequestInterface
    - Initial queryparameter check
    - KeywordStore usage
    - (Unused) controller router import
    - (Now unneeded) controller model import

Changes:
    - getQuery changes:
        - getQuery has been renamed to getAgendaItems
        - getAgendaItems is now passed through export & import
        - Simplified variable usage
  - Added Route.routeController property to clean up code
  - Empty queryParam inputs now also get removed from the url
  - agendaItems & currentPage are now passed through controller.set, eliminating need for a controller.setup


LBONTEN (commented):
That looks a lot cleaner. But it currently conflicts with master so we will need to fix that before resolving this PR

aliokan:
@Denperidge When I test locally, the infiniteScroll is broken. As I can see there is difference between filter in the 1st and the 2nd call: 
## initial call
image

## scroll call
image

Denperidge:
> @Denperidge When I test locally, the infiniteScroll is broken. As I can see there is difference between filter in the 1st and the 2nd call:
> ## initial call
> 
> image
> ## scroll call
> 
> image

This should now have been fixed, thanks for finding that bug! I've now cleaned it up even more

app/routes/agenda-items.ts
```diff
@@ -65,42 +64,69 @@ const getQuery = ({
...
-        };
-      };
-    };
-  };
+export async function getAgendaItems(
+  context: AgendaItemsRoute | AgendaItemsController,
+  params: AgendaItemsParams,
+  currentPage = 0
+) {
+  const controller =
```
aliokan, line 72: **Nitpick** I feel like this method has too much on its shoulders. What about only keep the query call here, and moving out all code related to controller/context?

aliokan (approved):
I left one none blocking comment. Feel free to merge. :+1:

app/routes/agenda-items.ts
Windvis: Agreed. I think this can be simplified a lot. It's a bit confusing to follow at the moment. I took the opportunity to experiment with "Resources" a bit, so I'll PR something soon that should improve that a bit.

PR #76, opened 2023-07-25
Fix detail async data display

aliokan:
This PR fix missing data loading in detail page. 

## Context

In order to fix detail page render when resolution value has Literal type, include parameter was removed from queries in detail page. 
https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/68/files#diff-071a3020b5ae0cb90e61e65c431b8d30c2ea445f40568cdbce2bd769a84bf6faL19-L30

As a side effect, all relationships expected to be sync are async now and not display correctly. 

## Before

image

## After

image


LBONTEN (approved):
Beautiful!

Denperidge (approved):
Looks good! Merge whenever you want

PR #77, opened 2023-07-26
Fix display all voters

aliokan:
Display all voters in Detail page "Wie stemde?" table. 

## Context

In order to fix detail page render when resolution value has Literal type, include parameter was removed from queries in detail page. 
https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/68/files#diff-071a3020b5ae0cb90e61e65c431b8d30c2ea445f40568cdbce2bd769a84bf6faL19-L30

Now all relations are loaded asynchronously.

As a side effect, default pagination (20) is applied on voters (`hasProponents', 'hasOpponents', 'hasAbstainers`). There is no pagination in include data. 

## Proposale

Increase page size to 100 on vote `hasProponents', 'hasOpponents', 'hasAbstainers` relationships. 

## Before

image

## After 

image

### DIFF
--- a/app/adapters/vote.ts
+++ b/app/adapters/vote.ts
@@ -1,0 +1,23 @@
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import Store, { Snapshot } from '@ember-data/store';
+
+export default class VoteAdapter extends JSONAPIAdapter {
+  findHasMany(
+    store: Store,
+    snapshot: Snapshot,
+    url: string,
+    relationship: { key: string }
+  ) {
+    // add page size 100 when relationship is voters
+    if (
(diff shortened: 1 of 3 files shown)


app/adapters/vote.ts
Denperidge, line 1: I wanted to ask if it perhaps would be easier/more in line with our other requests to have the page size set on the request, but I guess this is mostly used through relationships; in which adapters make more sense!

Denperidge (approved):
LGTM!

app/adapters/vote.ts
aliokan: Here it is ember-data which launches the request. I prefer to keep it like this instead of loading all voters imperatively into a controller.

PR #80, opened 2023-07-31
Embroider config tweaks

Windvis:
A new Embroider update came out which introduced some new config options which are interesting for our use case.

## `staticEmberSource`
`staticEmberSource` specifically has the potential to lower the bundle size once again, "for free".

Some quick `npm run build` comparisons showed the following results (I skipped the files that didn't change):

```
staticEmberSource: false
 - dist/assets/chunk.0add47c085d9ea410eef.js: 7.21 kB (1.41 kB gzipped)
 - dist/assets/chunk.99e2edea6ede5f06e637.js: 604.6 kB (142.34 kB gzipped)
 - dist/assets/vendor.90e0ebe1ad8dd88a12e3390b6fe2cf1f.js: 615.69 kB (177.35 kB gzipped)

staticEmberSource: true
 - dist/assets/chunk.d604db23e6947b8fd934.js: 3.13 kB (940 B gzipped)
 - dist/assets/chunk.ad47b1e614a2b431bc6f.js: 956.34 kB (249.72 kB gzipped)
 - dist/assets/vendor.ae41f34d9440502426bb299f92414ed9.js: 183.79 kB (57.37 kB gzipped)
 ```

Most of the code simply moved from vendor to a chunk but it's still a decrease of 84,24 kB (13,08 kb gzipped) overal. I'll take it.

## `amdCompatibility`
`amdCompatibility` only has a tiny influence on the size (+- 1kB smaller) but it's a prerequisite for the Vite bundler that is currently in development for Embroider so we might as well enable it already. It makes Embroider more strict which highlights when addons are doing strange things with dependencies.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -18,6 +18,13 @@ module.exports = function (defaults) {
     staticModifiers: true,
     staticComponents: true,
     staticEmberSource: true,
+    amdCompatibility: {
+      es: [
+        // au-date-range-picker imports jquery but doesn't depend on the jquery package directly. Instead it uses `@ember/jquery` which Embroider can't detect.
+        // We can either update the addon to depend on jquery directly (+ ember-auto-import) or switch to a different solution that doesn't use jquery at all
+        ['jquery', ['default']],
+      ],
+    },
 
     // This config is needed to work around an Ember Data v4.12 + Embroider issue
     // We can remove it once the fix is released: // https://github.com/embroider-build/embroider/issues/1506

### CODE REVIEW

File: ember-cli-build.js [hunk 1]
```diff
@@ -18,48 +17,17 @@ module.exports = function (defaults) {
     staticHelpers: true,
     staticModifiers: true,
     staticComponents: true,
-    splitAtRoutes: ['map'],
```
Windvis, line 21: It seems the route was removed so we no longer need this config either.

aliokan (approved):
Look good, tested locally, thanks! :pray:

PR #85, opened 2023-08-02
Only enable plausible when the config is provided

Windvis:
This prevents 405 calls because of the missing domain and host configuration on the dev environment.

The setup is similar to the one in loket: https://github.com/lblod/frontend-loket/commit/8d07d41c7f0d2a89c8ff0155f8727c87e07fbf9b

### DIFF
--- a/types/ember-plausible/services/plausible.d.ts
+++ b/types/ember-plausible/services/plausible.d.ts
@@ -1,0 +1,27 @@
+declare module 'ember-plausible/services/plausible' {
+  import type Service from '@ember/service';
+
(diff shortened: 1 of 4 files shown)


types/ember-plausible/services/plausible.d.ts
```diff
@@ -0,0 +1,27 @@
+declare module 'ember-plausible/services/plausible' {
```
Windvis, line 1: ember-plausible doesn't export types yet, so I've added some basic ones here for now.

PR #87, opened 2023-08-03
Use a resource to load the agenda items

Windvis:
This refactors the agenda items data loading to use a resource. This has some benefits:
- most of the logic is contained in 1 place
- the route and controller code is simplified a lot

### DIFF
--- a/app/controllers/agenda-items/index.ts
+++ b/app/controllers/agenda-items/index.ts
@@ -1,28 +1,119 @@
 import Store from '@ember-data/store';
 import Controller from '@ember/controller';
-import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+import { Resource } from 'ember-resources';
 import MunicipalityListService from 'frontend-burgernabije-besluitendatabank/services/municipality-list';
 import AgendaItem from 'frontend-burgernabije-besluitendatabank/models/agenda-item';
 import {
-  getAgendaItems,
-  AgendaItemsParams,
-} from 'frontend-burgernabije-besluitendatabank/routes/agenda-items';
+  AdapterPopulatedRecordArrayWithMeta,
+  getCount,
+} from 'frontend-burgernabije-besluitendatabank/utils/ember-data';
+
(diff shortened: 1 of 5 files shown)


app/controllers/agenda-items/index.ts
```diff
@@ -46,34 +145,55 @@ export default class AgendaItemsIndexController extends Controller {
...
+  plannedStartMax?: string;
+}) => ({
+  include: [
+    'sessions.governing-body.is-time-specialization-of.administrative-unit.location',
+    'sessions.governing-body.administrative-unit.location',
+  ].join(','),
+  sort: '-sessions.planned-start',
+  filter: {
+    sessions: {
+      ':gt:planned-start': plannedStartMin ? plannedStartMin : undefined,
```
Windvis, line 178: It would probably be better to simply not add the filters if the value isn't set instead of using `undefined`. I'm not sure if this would mean less work for mu-cl-resources, but if it does it potentially increases the speed. Since mu-auth is coming it might not be worth tinkering with it though.

app/routes/agenda-items/index.ts
```diff
@@ -1,172 +1,18 @@
...
-  controller.set('count', getCount(agendaItems));
-
-  return agendaItems.slice();
-}
-
-export default class AgendaItemsIndexRoute extends Route {
-  @service declare store: Store;
-  @service declare municipalityList: MunicipalityListService;
 
+export default class AgendaItemsRoute extends Route {
```
Windvis, line 3: Whoops, seems I made a mistake during the merge conflict resolving. Shouldn't have been renamed back.

app/controllers/agenda-items/index.ts
aliokan: When value is `undefined`, it is not part of the request. I guess it's filtered by `ember-data`.

aliokan (approved):
much cleaner! Tested locally

app/controllers/agenda-items/index.ts
Windvis: TIL 👍 I could sworn I've seen it output `undefined` somewhere in requests before 😄.

PR #89, opened 2023-08-07
Some more bundle size improvements

Windvis:
This adds a bundle analyzer setup (based on the [Embroider guide](https://github.com/embroider-build/embroider/blob/main/docs/analyzing.md#analyzing-bundles)) and adds some small bundle size improvements as well.


app/helpers/iso-formatter.ts
```diff
@@ -1,4 +1,4 @@
-import { format, parseISO } from 'date-fns';
+import { lightFormat, parseISO } from 'date-fns';
```
Windvis, line 1: The output of `lightFormat` is the same in our use-cases, and it's a lot smaller (3KB vs 24KB minified).

package.json
```diff
@@ -87,7 +88,7 @@
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-sass": "^11.0.1",
     "ember-cli-typescript": "^5.2.1",
-    "ember-concurrency": "^3.0.0",
+    "ember-concurrency": "^3.1.0",
```
Windvis, line 91: This forced a dedup of ember-concurrency, so we only ship v3 now (30Kb less).

app/helpers/iso-formatter.ts
Windvis, line 1: I don't think this file is used anymore, so we can probably just remove it?

aliokan: @Windvis Yes! :wastebasket:

aliokan (approved):

app/helpers/iso-formatter.ts
Windvis: I'll do it in a follow-up PR, I believe there were some other things that were unused as well.

PR #92, opened 2023-08-08
Only enable the `staticEmberSources` and `amdCompatibility` options in production builds

Windvis:
When `staticEmberSource` flag is enabled, the ember source modules are no longer exposed in the AMD system. This causes issues in the Ember inspector since it assumes certain modules are available that way.

As a workaround we only enable those options in production builds. This allows us to use the dev tools for local development, but still receive the bundle size improvements in production builds.


aliokan (approved):
Thanks! :pray:

PR #94, opened 2023-08-10
Fix the select filters on mobile devices

Windvis:
Due to the focus trap it wasn't possible to use the select filters properly on mobile devices. By adding the dropdown container to the allowed list, everything works as expected again.

### DIFF
--- a/app/templates/agenda-items/index.hbs
+++ b/app/templates/agenda-items/index.hbs
@@ -78,12 +78,13 @@
           {{/if}}
         {{/let}}
       </div>
-      <FilterSidebar 
-        @id="filtersidebar" 
+      <FilterSidebar
+        @id="filtersidebar"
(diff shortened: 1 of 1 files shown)


app/templates/agenda-items/index.hbs
```diff
@@ -78,12 +78,13 @@
...
       </div>
-      <FilterSidebar 
-        @id="filtersidebar" 
+      <FilterSidebar
+        @id="filtersidebar"
         @class="c-interface__sidebar c-interface__sidebar--dialog {{if this.hasFilter 'is-visible'}}"
         {{focus-trap
           isActive=this.hasFilter
           shouldSelfFocus=true
+          additionalElements=(array "#ember-basic-dropdown-wormhole")
```
Windvis, line 87: The range picker doesn't work properly on mobile either, but that's being replaced by #88 so I didn't include that container here.

aliokan (approved):
:+1:

PR #96, opened 2023-08-11
Agenda item details page tweaks

Windvis:
- update the logic to determine when an agenda item is handled. We now use the `handledBy` relationship. If it is set, we consider the item handled.
- Always show the "vote" accordion even if the data is missing, or if the item wasn't handeld yet. If there is no vote data we display a message.

### DIFF
--- a/app/controllers/agenda-items/agenda-item.ts
+++ b/app/controllers/agenda-items/agenda-item.ts
@@ -12,6 +12,10 @@ export default class AgendaItemController extends Controller {
     return !!this.model.vote;
   }
 
+  get wasHandled() {
+    return Boolean(this.model.agendaItem.belongsTo('handledBy').value());
+  }
+
   get municipalityQuery() {
     return { gemeentes: this.model.agendaItem.session?.municipality };
   }
--- a/app/templates/agenda-items/agenda-item.hbs
+++ b/app/templates/agenda-items/agenda-item.hbs
@@ -21,7 +21,7 @@
     <div class="au-o-box">
       <div class="au-o-grid au-o-grid--small">
         <div class="au-o-grid__item au-u-hidden-from@medium">
-          {{#if this.model.agendaItem.handledBy.resolutions}}
+          {{#if this.wasHandled}}
             <AuAlert @skin="success" @icon="decided" @title="Dit werd behandeld" class="au-u-margin-bottom-none" />
           {{else}}
             <AuAlert @skin="warning" @icon="three-dots" @title="Dit werd nog niet behandeld" class="au-u-margin-bottom-none" />
@@ -106,7 +106,7 @@
                   {{else}}
                     <AuAlert
                       @skin="warning"
-                      @title="Geen verslag"
+                      @title={{if this.wasHandled "Geen verslag" "Dit agenda punt werd nog niet behandeld"}}
                       class="au-u-margin-top au-u-margin-bottom"
                     />
                   {{/if}}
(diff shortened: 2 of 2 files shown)


app/controllers/agenda-items/agenda-item.ts
```diff
@@ -12,6 +12,10 @@ export default class AgendaItemController extends Controller {
     return !!this.model.vote;
   }
 
+  get wasHandled() {
+    return Boolean(this.model.agendaItem.belongsTo('handledBy').value());
```
Windvis, line 16: This assumes the relationship is loaded (which is why I added the include). We _could_ do it from the template but it would require `{{#let}}` and some sort of `await` helper combo that wraps all the places where we need it. I prefer this approach myself since it makes the transition to sync relationships easier (which I still think is the best setup).

app/templates/agenda-items/agenda-item.hbs
```diff
@@ -106,30 +106,40 @@
...
+              <section class="c-accordion-holder">
+                <AuBadge @icon="vote" @skin="brand" />
+                <AuAccordion
+                  @subtitle={{this.model.vote.subject}}
+                  @reverse={{true}}
+                  @iconOpen="nav-down"
+                  @iconClosed="nav-right"
+                  @buttonLabel="Stemming"
+                >
+                  {{#if this.hasVote}}
```
Windvis, line 126: typo..

app/routes/agenda-items/agenda-item.ts
```diff
@@ -13,7 +13,9 @@ export default class AgendaItemRoute extends Route {
   @service declare keywordStore: KeywordStoreService;
 
   async model(params: DetailParams) {
-    const agendaItem = await this.store.findRecord('agenda-item', params.id);
+    const agendaItem = await this.store.findRecord('agenda-item', params.id, {
+      include: 'handled-by',
+    });
```
aliokan, line 18: **Issue:** The include is not needed. We already load it line 53: `const agendaItemHandling = await agendaItem.handledBy;`
We decide to split all require in small calls, in order to improve source of problem and avoid on big 500 error when a property or a link is broken.   

```suggestion
   // To improve error isolation, the required data are loaded below in separate calls. 
    const agendaItem = await this.store.findRecord('agenda-item', params.id);
```

app/routes/agenda-items/agenda-item.ts
```diff
@@ -13,7 +13,9 @@ export default class AgendaItemRoute extends Route {
...
 
     // wait until sessions are loaded
     const sessions = await agendaItem.sessions;
```
aliokan, line 21: I put the comment here, but concern line 54. `const vote = (await agendaItemHandling?.hasVotes)?.toArray().shift();` is not needed here anymore.

aliokan (changes requested):
I left one comment regarding the require in agendaitem call. :ping_pong:

app/routes/agenda-items/agenda-item.ts
Windvis: Oh, I missed that. I saw a `handled-by` call in the network tab and assumed it was done by the template access. Thanks. I'm not sure I understand the reasonings behind the split, but I'll just change it for now.

aliokan: @Windvis Before, we used require, but when something was broken in the data (like a type error in besluit value), the server returned a 500 error. The whole page was empty, and it was difficult to isolate the source of the problem. Now each dependency is loaded separately. We render the page even when some data is broken, and it is easier to find which data is broken.

app/routes/agenda-items/agenda-item.ts
Windvis: Maybe I misunderstand by `this.model.vote` is used in the `hasVotes` getter in the controller, so I think it's still needed?

aliokan: If instead we use `this.model.vote` directly in template should it load data when we open the accordion?

Windvis: Yea, but I really dislike loading data through the template. It makes things hard to follow. If you don't want to load it in the model hook I would use a resource instead. That would also allow us to display loading states, instead of the glitchy rendering that happens otherwise.

It also seems unrelated to what this PR is doing, so I think we can do that in a seperate one? (regardless if we load through the template or through a different setup).

aliokan: :+1:

app/routes/agenda-items/agenda-item.ts
Windvis: (I force pushed the revert, so a reset is needed if you want to test again)

aliokan (approved):
LGTM! :+1:

PR #99, opened 2023-08-11
Remove a rogue .DS_Store file

Windvis:
This was probably already checked in before it was added to the .gitignore.


aliokan (approved):
Thanks!

PR #100, opened 2023-08-14
Refactor the `separator` constant

Windvis:
Extracting the logic to some utils makes the separator an implementation detail so it's easier to change in the future.

### DIFF
--- a/app/components/filters/filter.ts
+++ b/app/components/filters/filter.ts
@@ -1,16 +1,17 @@
 import Component from '@glimmer/component';
 import RouterService from '@ember/routing/router-service';
 import { service } from '@ember/service';
-import { get } from '@ember/object';
+
+export interface FilterArgs {
+  id: string;
+  queryParam: string;
+
+  searchField: string;
+  options: Array<object>;
+}
 
 interface Signature {
-  Args: {
-    id: string;
-    queryParam: string;
-
-    searchField: string;
-    options: Array<object>;
-  };
+  Args: FilterArgs;
 }
 
 export default class FilterComponent<S = Signature> extends Component<S> {
@@ -21,8 +22,8 @@ export default class FilterComponent<S = Signature> extends Component<S> {
    * @param param name of the queryParameter that is presented to the user
    * @returns queryParameter value. Possibly undefined
    */
-  getQueryParam(param: string) {
-    return get(this.router.currentRoute.queryParams, param);
+  getQueryParam(param: string): string | undefined {
+    return this.router.currentRoute.queryParams[param];
   }
 
   /**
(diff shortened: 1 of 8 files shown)


app/components/filters/select-multiple-filter.ts
```diff
@@ -37,19 +46,15 @@ export default class SelectMultipleFilterComponent extends FilterComponent {
...
-    values = values.replace(regex, '');
+  async selectChange(selectedOptions: Option[]) {
+    this.selected = selectedOptions;
 
     this.updateQueryParams({
-      [this.args.queryParam]: values,
+      [this.args.queryParam]: serializeArray(
+        selectedOptions
+          .map((value) => value[this.args.searchField])
+          .filter(Boolean) as string[]
```
Windvis, line 56: `as string[]` is needed because TS doesn't narrow the type to thruthy only values while in practice it can't ever be false after the `.filter`.

app/services/municipality-list.ts
```diff
@@ -54,11 +54,10 @@ export default class MunicipalityListService extends Service {
    * @returns a Promise for a joined string of those locations' id's, or undefined
    */
   async getLocationIdsFromLabels(
-    labels?: Array<string> | string,
-    stringSeperator = seperator
```
Windvis, line 58: We didn't pass custom separotors anywhere so I simply removed it. We could add an option to the serialize / deserialize utils to configure the separator, but I don't think it's needed.

app/templates/agenda-items/index.hbs
```diff
@@ -102,7 +102,6 @@
             @searchField="label"
             @queryParam="gemeentes"
             @placeholder="Alle gemeentes"
-            @value={{this.municipality}}
```
Windvis, line 105: The value wasn't used since it is retrieved from the router service.

app/components/filters/select-multiple-filter.ts
aliokan: Yeah, I have to look for the syntax every time I use a filter ^^. I don't know why they can't handle it correctly after all these years

app/components/filters/filter.ts
```diff
@@ -21,8 +22,8 @@ export default class FilterComponent<S = Signature> extends Component<S> {
    * @param param name of the queryParameter that is presented to the user
    * @returns queryParameter value. Possibly undefined
    */
-  getQueryParam(param: string) {
-    return get(this.router.currentRoute.queryParams, param);
+  getQueryParam(param: string): string | undefined {
```
aliokan, line 25: **Question:** Any reason to add the type here? Is the type inference broken?

app/templates/agenda-items/index.hbs
aliokan: I guess this kind of issues will be caught with Glint template

aliokan (approved):
Look good to me, can be merged!

app/components/filters/filter.ts
Windvis: No, just a personal preference for explicit return types, but I can remove it.

Windvis: I found a post that describes pretty well why I prefer explicit return types: https://stackoverflow.com/a/70001850

PR #102, opened 2023-08-14
Set up Stylelint for Sass files

Windvis:
This adds stylelint-config-standard-scss and resolves all linting issues.

Based on https://github.com/lblod/frontend-lpdc/pull/6


package.json
```diff
@@ -69,6 +62,13 @@
     "@types/ember__test": "^4.0.1",
     "@types/ember__test-helpers": "^2.9.1",
     "@types/ember__utils": "^4.0.2",
+    "@types/ember-data": "^4.4.11",
```
Windvis, line 65: I don't understand why this moves around all the time 😄 seems like a mismatch between npm versions maybe?

aliokan (approved):
I like this kind of improvement!

PR #103, opened 2023-08-16
feat: :sparkles: added a new hasVotes in the vote component

snenenenenenene:
- added a new hasVotes get in votes as I thought it'd be more consolidated in the vote model
- kept the rest of the hasVotes gets in case I stepped on toes

### DIFF
--- a/app/components/vote-overview.hbs
+++ b/app/components/vote-overview.hbs
@@ -54,83 +54,81 @@ not triggering the linting error}}
   </div>
 {{/if}}
 
-{{#if this.model.vote.hasVoteNumbers}}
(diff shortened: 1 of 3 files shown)


aliokan (changes requested):
As discuss during standup, we want to still display the vote when there is not numbers.

snenenenenenene:
@aliokan fixed it!

aliokan (changes requested):
Changes are good, I still have a blocker regarding hidden voters table based on `hasVoteNumbers` value

snenenenenenene:
Fixed it @aliokan

aliokan (approved):
The modification looks good. LGTM! :pray:

PR #105, opened 2023-08-23
fix: Load agendaItems without governingBody `is-time-specialization-of`

aliokan:
## Context

Some municipalities do not define a Governing-body with temporal specialization. In this case, the filtering of the municipality does not return a result.

## Proposal

~~When filtering on municipality by going through `sessions.governing-body.is-time-specialization-of.administrative-unit.location` don't return result. Try to filter without time specialization : `sessions.governing-body.administrative-unit.location`.~~

- Use `filter[:or:]` to filter on municipality with or without Governing-body with temporal specialization.
- Use `filter: keyword` instead of `filter[:or:][title]` and `filter[:or:][description]` to filter on keyword.

This solution is better than the previous one, because it handles municipality that mix Governing-body with and without temporal specialization like Gent, Genk or Brugge. 

## Before

image

## After

image

### DIFF
--- a/app/controllers/agenda-items/index.ts
+++ b/app/controllers/agenda-items/index.ts
@@ -74,7 +74,7 @@ class AgendaItemsLoader extends Resource<AgendaItemsLoaderArgs> {
 
       const { keyword, plannedStartMin, plannedStartMax } = this.#filters;
 
-      const agendaItems: AdapterPopulatedRecordArrayWithMeta<AgendaItem> =
+      let agendaItems: AdapterPopulatedRecordArrayWithMeta<AgendaItem> =
         await this.store.query(
           'agenda-item',
           agendaItemsQuery({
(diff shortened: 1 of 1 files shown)


snenenenenenene:
I'm having a problem running this branch properly.

When I run `npm run dev:proxy` on this branch I don't get any agenda items whatsoever.
image

When I switch back to master this works properly. 
image

app/controllers/agenda-items/index.ts
```diff
@@ -173,25 +173,13 @@ const agendaItemsQuery = ({
...
-        },
-      },
-    },
-    ':or:': {
-      title: keyword ? keyword : undefined,
-      description: keyword ? keyword : undefined,
-    },
-  },
+  filter: keyword || undefined,
+  'filter[:or:][sessions][governing-body][is-time-specialization-of][administrative-unit][location][:id:]':
```
aliokan, line 177: Complex filters are to be string keys because we cannot mix: `filter: keyword || undefined,` and `filter: {` in the same object.

aliokan:
@snenenenenenene A new solution is available now, can you take a look?

aliokan:
Merge in order to move forward

PR #107, opened 2023-08-30
fix: :bug: overflow fix on mobile devices

snenenenenenene:
- changes css classes to fix overflow issue

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -267,3 +267,25 @@ cursor {
     max-width: map.get($mq-breakpoints, "xsmall");
   }
 }
+
+.au-c-main-container__content {
+  flex-grow: 999;
+  flex-basis: 0;
+  height: 100%;
+  max-width: 100%;
+}
+
+.au-c-app {
+  overflow: hidden;
+}
+
+.au-c-app .au-c-main-container {
+  height: auto;
+  overflow: hidden;
+}
+
+.c-interface {
+  flex-direction: row-reverse;
+  align-items: stretch;
+  height: 100%;
+}


snenenenenenene:
@aliokan one issue however is that the footer is gianormous, and i'm not so certain my aunty woud love to look at a giant footer when she's scrolling agenda items.

image

aliokan:
@snenenenenenene InfiniteScroll and footer are definitely not friends. Does this footer really make sense on mobile? What do you think to move the footer into a burger menu on mobile?

@Windvis Do you have any idea here?

Windvis:
> one issue however is that the footer is gianormous

It seems they tweak the styles a bit for mobile devices on the vlaanderen.be website:

image

> InfiniteScroll and footer are definitely not friends.

Indeed, that's why I marked it as a low prio bug. I think clicking a button is better than not being able to see the footer. I don't know the solution to this either. Usually layouts with infinite scrolling don't use footers, but I'm not sure if there are flanders-themed examples in the wild that do this already.

Maybe the UX team has some advice?

Windvis:
Also, I would _really_ urge not to mess with the css and simply use a non-scroll based solution. That would also work on very large (high) screens for example. With the addons linked in the issue that shouldn't even be that hard either.

snenenenenenene:
> > one issue however is that the footer is gianormous
> 
> It seems they tweak the styles a bit for mobile devices on the vlaanderen.be website:
> 
> image
> 
> > InfiniteScroll and footer are definitely not friends.
> 
> Indeed, that's why I marked it as a low prio bug. I think clicking a button is better than not being able to see the footer. I don't know the solution to this either. Usually layouts with infinite scrolling don't use footers, but I'm not sure if there are flanders-themed examples in the wild that do this already.
> 
> Maybe the UX team has some advice?

I asked Saska and she gave me these options:

1. Adding pagination and having the footer at the bottom of the page (not screen)
2. Writing it so that we know what phone the user uses, and always show the footer "under the fold" on a phone (or partially, for example, always being able to see the Vlaanderen logo, that way the user can scroll through the results, and if they wish to look at the footer, the can scroll to the footer grabbing the logo and scrolling that way)
3. Rethinking the app-universe footer: e.g. adding the three menu items into a little drop-down to save a couple of pixels in height. having a collapsible footer with a dropdown on the same level as the logo to show the rest,... This option might need someone's approval (not sure whose) as it changes an element of AU that is inherently ABB.

PR #108, opened 2023-08-30
Added help page

snenenenenenene:
## DRAFT

- Added FAQ component
- Added help route
- Added FAQ from Julie's document

One question I have is that I added expandable cards, but I'm not so sure we'll be using those...
We can either chose to give a short explanation in the visible card and move the nitty gritty to the expandable part OR move all the information into the visible card part.
Would love to have your oppinion!!!

### DIFF
--- a/app/components/faq-card.hbs
+++ b/app/components/faq-card.hbs
@@ -2,7 +2,7 @@
   id="help-zoekresultaten"
   @expandable={{true}}
   @flex={{true}}
-  @divided={{true}}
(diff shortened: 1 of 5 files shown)


snenenenenenene:
## Option 1:
(think of it without the expand arrow)

image

## Option 2:

image

image

snenenenenenene:
@aliokan updated the card

aliokan (approved):
Great! :+1:

PR #109, opened 2023-08-30
fix(agendaItem): Add fallback when no municipality in session

aliokan:
Select at least one session, even if there is no linked municipality. 

## Before

image

## After

image


aliokan:
close because woodpecker error on feature branch. Open a new one: https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/110

PR #110, opened 2023-08-30
fix(agendaItem): Add fallback when no municipality in session

aliokan:
Select at least one session, even if there is no linked municipality. 

## Before

image

## After

image


snenenenenenene:
Functions perfectly!
However, I was wondering whether we should always include the bestuurseenheid/naam if the data does not contain it.

aliokan:
I created a ticket, to completely removed these agenda-items: https://binnenland.atlassian.net/browse/BNB-319

PR #118, opened 2023-09-13
Reduce footer size

Dietr:
Reduces the overall footer size on all layouts.
Shows the footer on scroll on the detail layout.


Dietr:
I moved the footer on the detail page as it works with the scroll there an there's no need to have it fixed at the bottom. On the list page I can't move it because of the infinite scroll. 
Reasoning is to always show as much content as possible on the screen.

snenenenenenene:
Wouldn't it become a bit of a hassle for people to look at the footer on the detail page on mobile?
When the infinite scroll - without clicking the "load more" button on mobile - gets approved in one of the open PR's, won't it become close to impossible to look at the footer?

aliokan (approved):
Move forward, footer will be improve in the future

PR #121, opened 2023-09-15
fix(agendaItems): Support mu-search array of date

aliokan:
- add utils/mu-search-data-format/parseDateStringToDate
- use parseDateStringToDate in dataMapping

## Before

image

## After

image


snenenenenenene (approved):
Looks good to me!

PR #131, opened 2023-09-22
fix: Agenda item controller

aliokan:


aliokan:
@LBONTEN template and controller need to follow the same path struture

aliokan:
hotfix: Merge and deploy to fix detail page

PR #133, opened 2023-09-22
fix: Remove remained `not handeled` alert

aliokan:
Fix mobile view

Before

image

After

image


LBONTEN (approved):
Looks like I missed this one yesterday

PR #134, opened 2023-09-22
moved function to controller

LBONTEN:
Moved the code to get the first four agenda items of a session from the route to the controller. 
This way I was able to reuse already fetched data and keep logic in the controller.

### DIFF
--- a/app/controllers/agenda-items/agenda-item/session.ts
+++ b/app/controllers/agenda-items/agenda-item/session.ts
@@ -1,7 +1,6 @@
 import Controller from '@ember/controller';
 import { ModelFrom } from '../../../lib/type-utils';
 import { sortObjectsByTitle } from 'frontend-burgernabije-besluitendatabank/utils/array-utils';
-import AgendaItem from 'frontend-burgernabije-besluitendatabank/models/agenda-item';
 import AgendaItemsAgendaItemSessionRoute from 'frontend-burgernabije-besluitendatabank/routes/agenda-items/agenda-item/session';
 
 export default class AgendaItemsAgendaItemSessionController extends Controller {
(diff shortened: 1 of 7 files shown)


app/controllers/agenda-items/agenda-item/session.ts
```diff
@@ -1,22 +1,16 @@
...
-      | AgendaItem[]
-      | null;
-
-    if (!agendaItems) {
-      agendaItems = [];
-    }
-
-    return agendaItems.slice().sort(sortObjectsByTitle);
+    return this.model.agendaItemOnSameSession
+      ?.concat(this.model.agendaItem)
```
aliokan, line 12: **Issue:**  `concat` create a duplicate entry
image

aliokan (changes requested):
@LBONTEN During testing, I notice a difference result when a session contains more than 20 agenda-items:  
- `sessions/:session_id`, all agenda-items are visible; 
- `agenda-items/agenda-item/session`, only 20 agenda-items are visible. 

For example `zittingen/597f1df0-53d0-11ee-b56c-e741b3e2335a` on dev server. 

## Proposal

I see two options here: 
- keep request on `session` in `app/controllers/agenda-items/agenda-item/session.ts`, there is no pagination on `include` content. 
- or increase the page size in `agendaItemOnSameSession` request.

app/controllers/agenda-items/agenda-item/session.ts
LBONTEN: Oh strange. I added the concat because it seemed the 'original' agenda item was missing. But that must be because the filtering is now in the controller. 
Will be fixed :)

LBONTEN:
@aliokan reusing the request is already what I am doing in the previous implementation. Starting to think I'm just working backwards in this PR. So I think closing it would be best.

PR #138, opened 2023-09-26
feat: :sparkles: added the feedback widget

snenenenenenene:
Added feedback widget that link users to the jira feedback page.

### DIFF
--- a/app/components/feedback-widget.hbs
+++ b/app/components/feedback-widget.hbs
@@ -1,0 +1,3 @@
+<a class="feedback-widget" href="">
+  Feedback
+</a>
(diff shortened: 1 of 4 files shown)


aliokan:
@snenenenenenene Have you test on Chrome, Firefox, Safari, android and ios?

LBONTEN:
The button needed to say “Heb je opmerkingen, vragen of feedback?” love. Also, since it's just a temporary feature for feedback. Should it be merged into master? Can't we keep this separate. So it's easier to revert later.

aliokan (approved):
Move forward

PR #140, opened 2023-09-28
feat: :sparkles: added search to home page

snenenenenenene:


app/controllers/home.ts
```diff
@@ -28,11 +28,26 @@ export default class HomeController extends Controller {
...
+  }
+
+  @action
+  handleChange(event: Event) {
+    this.value = (event.target as HTMLInputElement).value;
+  }
 
   @action handleMunicipalitySelect() {
     this.loading = true;
+    console.log(this.value);
```
aliokan, line 47: **Issue:** @snenenenenenene oups ;)

app/controllers/home.ts
```diff
@@ -28,11 +28,26 @@ export default class HomeController extends Controller {
   ) {
     this.selectedMunicipalities = selectedMunicipalities;
   }
+  /** Handles keyword search on homepage */
+  @tracked value = '';
```
aliokan, line 32: **Nitpick:** Using a more meaningful name would improve readability and avoid confusion if other input text is added in the future 🙃

aliokan (changes requested):
@snenenenenenene I first thought that it was worth creating a component to encapsulate the input logic. But after all, there is a good chance that the input keyword on home and list page will have different behavior or design in the future. 

Except the remaining `console.log` we are good to move forward. 

Good job! :+1:

snenenenenenene:
@aliokan, fixed it! Thanks for the feedback <3

aliokan (approved):
Tested locally, looks good! :+1: 

LGTM! :rocket:

PR #147, opened 2023-10-11
Fix detail page similar itmes

aliokan:
Before

image

After

image


aliokan:
@snenenenenenene Now, the current agend-item is filtered out.

snenenenenenene (approved):
Looking great!!!!!!

PR #148, opened 2023-10-11
Daterange update

snenenenenenene:
- Added error state that notifies users when the start date exceeds the end date.
- Added (proprietary) min and max dates

Feel free to leave feedback on how to make it more userfriendly.
First I went with a blur handler that detects when a user clicks outside of the inputs to then validate the dates. But this extra click felt a bit unnatural to me.


aliokan (changes requested):
@snenenenenenene We have to block the submition when values are wrong. Currently, requests are made on each input change, even when the values are wrong. 

image

I thing that add a submit button will help with the validation and error handling.

snenenenenenene:
@aliokan, after talking to Jodi he said he'd prefer no submit button.
I ended up adding some error handling.

app/components/filters/date-range-filter.ts
```diff
@@ -200,9 +202,36 @@ export default class DateRangeFilterComponent extends Component<Signature> {
     }
   }
 
+  errorMessages = A<string>([]);
+
   updateQueryParamsIfValid() {
+    // if the start is before 2015 or the end after today + 100 years
     if (this.hasBothDates || this.hasNoDates) {
+      if (this.start && new Date(this.start) < new Date('2015-01-01')) {
```
aliokan, line 210: @snenenenenenene **Blocker:** start and end dates are different in class and template

aliokan (dismissed):
@snenenenenenene I have played a bit with the component and the errors are not reset correctly. Old errors still visible. 

image

app/components/filters/date-range-filter.ts
snenenenenenene: @aliokan I don't completely understand the problem, could you elaborate?

aliokan: @snenenenenenene In `app/components/filters/date-range-filter.ts` the date range is defined from `2015-01-01` to `2100-01-01`, but in `app/components/filters/date-range-filter/date-input.hbs` is from `2000-01-01` to `2099-12-31`

snenenenenenene: @aliokan should be fixed!

LBONTEN (commented):
I'm not a big fan of where the errors pop up. It limits space on where we could put other filters, etc. I feel like the error messages should float on top of the DOM.

aliokan (changes requested):
@snenenenenenene **Blocker:** Usually toasters are used to display notifications, not input errors. With the current implementation, users cannot easily understand which input field is wrong.

image

**Suggestion:** 
- use `@error=true` on `AuInput` component the content is not valid
- use `AuHelpText` component to display error messages 
You can find example in OP project: https://github.com/lblod/frontend-organization-portal/blob/767751824a315ffff6ce54b01548ab6e2ac0fa99/app/templates/administrative-units/administrative-unit/local-involvements/edit.hbs#L81-L83

Here an other example: 
image

snenenenenenene:
@aliokan, I've updated the logic!

aliokan (changes requested):
@snenenenenenene I still have these odd behaviors: 
- When remove the value of start or end input date, the URL is not updated (date not removed) due to `isDateComplete`; 
image

- When both inputs have errors, and you fix one, the error still visible. 
image

You can find a proposal to fix them: https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/156

Feel free to merge ;)

snenenenenenene:
@aliokan all should be good now!

aliokan (approved):
Works as expected :)

LGTM! :rocket:

PR #149, opened 2023-10-19
Feat/data-label

snenenenenenene:
- Added logic behind the data quality

### DIFF
--- a/app/controllers/agenda-items/agenda-item/index.ts
+++ b/app/controllers/agenda-items/agenda-item/index.ts
@@ -29,6 +29,7 @@ export default class AgendaItemController extends Controller {
 
   showModal = () => {
     this.modalOpen = true;
+    console.log(this.model.agendaItem.agendaItemQuality.qualityMetrics);
   };
 
   closeModal = () => {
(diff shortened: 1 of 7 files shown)


app/controllers/agenda-items/agenda-item/index.ts
```diff
@@ -29,6 +29,7 @@ export default class AgendaItemController extends Controller {
 
   showModal = () => {
     this.modalOpen = true;
+    console.log(this.model.agendaItem.agendaItemQuality.qualityMetrics);
```
aliokan, line 32: @snenenenenenene Is it needed?

app/models/agenda-item.ts
```diff
@@ -49,6 +49,61 @@ export default class AgendaItemModel extends Model {
   get dateFormatted() {
     return this.session?.dateFormatted;
   }
+
+  get agendaItemQuality(): {
+    qualityMetrics: { label: string; value: boolean }[];
+    qualityValue: number;
```
aliokan, line 55: @snenenenenenene Is it use somewhere?

aliokan (commented):
@snenenenenenene Except the 2 questions it's look good

app/controllers/agenda-items/agenda-item/index.ts
snenenenenenene: WOops, removed it!

app/models/agenda-item.ts
snenenenenenene: I've removed the quality value, it was a remnant of what was based on the mock ups. 
But it does seem we won't need it in the near future so I've deleted it!

PR #150, opened 2023-10-25
Style update

snenenenenenene:
- Changed colour of "Treated" from green to grey
- Fixed footer position


aliokan (approved):
Tested locally

image

PR #151, opened 2023-10-25
Statistics page

snenenenenenene:
- added statistics to the data quality page
- moved handledBy code from index to agenda item model
- added business logic to the data quality route
- moved the data quality badge up
- added a municipality filter to the data quality
- added loading functionality to the data quality


app/controllers/data-quality.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  get totalAgendaItemsWithoutTitleAndDescription(): number {
+    const model = this.model as AgendaItem[];
+    return model.filter((item: AgendaItem) => item.handledBy).length;
+  }
+
+  get topTenGoverningBodies(): number {
+    const model = this.model as AgendaItem[];
+    return model.filter((item: AgendaItem) => item.handledBy).length;
+  }
```
aliokan, line 77: **Blocker:** These methods doesn't seems to be used

app/routes/data-quality.ts
```diff
@@ -1,3 +1,191 @@
...
+      // resolve the promises
+      agendaItemsPerGoverningBodyClassification = await Promise.all(
+        agendaItemsPromises
+      );
+
+      agendaItemsPerGoverningBodyClassification =
+        agendaItemsPerGoverningBodyClassification.map((item) => ({
+          ...item,
+          percentage:
+            ((item.count || 0) / (getCount(totalCountAgendaItems) || 1)) * 50,
```
aliokan, line 143: **Nitpick:** `getCount(totalCountAgendaItems)` is executed 6 times. What about execute once and use the result everywhere?

app/routes/data-quality.ts
```diff
@@ -1,3 +1,191 @@
...
+      const agendaItemsWithVoteOrVoters: AdapterPopulatedRecordArrayWithMeta<VoteModel> =
+        await this.store.query('vote', {
+          'filter[:or:][handled-by][subject][sessions][governing-body][is-time-specialization-of][administrative-unit][location][label]':
+            params.municipalityLabel || undefined,
+          'filter[:or:][handled-by][subject][sessions][governing-body][administrative-unit][location][label]':
+            params.municipalityLabel || undefined,
+          size: 1,
+        });
+
+      return {
```
aliokan, line 163: **Nitpick:** In order to improve the readability. Can you keep the return object simple and execute all the computations before?

app/routes/data-quality.ts
```diff
@@ -1,3 +1,191 @@
...
+            100
+        ),
+        agendaItemsPerGoverningBodyClassification:
+          // first sort by label then sort by count descending set total at the begining of the array
+          agendaItemsPerGoverningBodyClassification
+            .sort((a, b) =>
+              (a.classification || '').localeCompare(b.classification || '')
+            )
+            .sort((a, b) => (b.count || 0) - (a.count || 0))
+            .sort((a, b) => (a.classification === 'Totaal' ? -1 : 0)),
```
aliokan, line 182: **Blocker:** Chaining `sort` calls will not have the effect that you describe in the comment. Here only the latest sort() will have effect. To do what it expected, you have to merge all code in one `sort` method. You can find an example of multicriterial sorting here: https://dev.to/markbdsouza/js-sort-an-array-of-objects-on-multiple-columns-keys-2bj1

aliokan (changes requested):
@snenenenenenene I left some comments ;)

app/routes/data-quality.ts
snenenenenenene: @aliokan I've resolved everything except for the sort. I can't get it to work the way it does currently using multicriterial sorting :/

aliokan (changes requested):
@snenenenenenene After this round, we should good to go :ping_pong:

app/routes/data-quality.ts
```diff
@@ -1,3 +1,197 @@
...
+      agendaItemsPerGoverningBodyClassification =
+        agendaItemsPerGoverningBodyClassification.map((item) => ({
+          ...item,
+          percentage: ((item.count || 0) / (totalCountAgendaItems || 1)) * 50,
+        }));
+
+      agendaItemsPerGoverningBodyClassification.push({
+        classification: 'Totaal',
+        count: totalCountAgendaItems || 0,
+        percentage: 50,
```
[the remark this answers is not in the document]
snenenenenenene: On smaller screens the graph gets buggy if it is at 100% width so I hardcode it at 50% so it's pretty much always clearly visible on classifications with long string lengths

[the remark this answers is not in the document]
aliokan: I prefer to keep data meaningful, and adapt data at the last place when it's needed. In other words, keep `percentage: 100` in data and divide the percentage in the template with a comment that explain why it's divided in 2. It will help the next developer or you in few weeks ;)

app/templates/data-quality.hbs
```diff
@@ -40,17 +40,6 @@
...
-          <div
-            class="au-u-flex au-u-flex--spaced-small au-u-flex--vertical-center"
-          >
-            <AuBadge @icon="dash" @skin="warning" @size="small" />
-            <p>
-              Er zijn documenten gevonden of meldingen ontvangen met dit
-              optioneel in te vullen attribuut
-            </p>
-          </div>
-        </Item>
```
aliokan, line 53: **Question:** Is it removed on purpose?

aliokan (approved):
@snenenenenenene There is no blocker anymore. I left some nitpick if you want polish a bit.

app/templates/data-quality.hbs
snenenenenenene: It has not been implemented, so I removed it :)

PR #156, opened 2023-11-03
Simplify daterange component

aliokan:
@snenenenenenene This a proposal to simplify the way that error handling is managed on the component. 

It fix 2 broken behaviors : 
- When remove the value of start or end input date, the URL is not updated (date not removed) due to `isDateComplete`; 
image

- When both inputs have errors, and you fix one, the error still visible. 
image


snenenenenenene (approved):
LGTM!!

PR #163, opened 2023-11-10
fix: :bug: added overflow class to safeguard small home component

snenenenenenene:


aliokan (approved):
Tested locally, it works

PR #167, opened 2023-11-15
Improve help and data-quality text

aliokan:
image

image


aliokan:
As it's only text, I force merge to create the new release

PR #172, opened 2023-11-21
fix: :bug: added quick modal fix in route

snenenenenenene:


app/routes/agenda-items/agenda-item.ts
```diff
@@ -13,6 +13,12 @@ export default class AgendaItemRoute extends Route {
   @service declare keywordStore: KeywordStoreService;
 
   async model(params: DetailParams) {
+    // eslint-disable-next-line ember/no-controller-access-in-routes
```
aliokan, line 16: **Question:** Can we find a solution without disable eslint ;)

Have you try to use `resetController` as describe here: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-controller-access-in-routes.md

snenenenenenene: I've tried it before, but I can't get the resetController hook to be fired. :/

`
  resetController(
    controller: AgendaItemController,
    isExiting: boolean,
    transition: Transition<unknown>
  ) {
    super.resetController(controller, isExiting, transition);
    if (isExiting) {
      controller.set('modalOpen', false);
      controller.modalOpen = false;
    }
  }
`

aliokan (approved):
LGTM! :)

PR #176, opened 2023-12-04
feat: :sparkles: province-filter

snenenenenenene:
After talking about it we made the decision to put all of the locations into one monolithic location search. This PR aims to add provinces to said search expanding on the currently implemented municipalities.
However, this not only means changing the select to work with multiple groups. It also means bringing the queryParam => select and vice versa up to speed.

- Changing the select to also use groups
- Changing the "insert" code to add selected options even in groups which use nested option arrays
- Changing the "selectOption" code to update variables of multiple query params as well as making it be able to append variables of the same type/groupName to their respective query parameters.


app/components/filters/select-multiple-filter.ts
```diff
@@ -1,29 +1,58 @@
...
+          (value) => get(value, searchField) === needle
+        );
+        if (found) {
+          results.push(found);
+        }
+      }
+
+      this.selected = results;
+    } else {
+      const queryParam = this.getQueryParam(this.args.queryParam) as string;
```
aliokan, line 54: **Blocker**: You do not handle the case, when `this.args.queryParam` is undefined.

app/components/filters/filter.ts
```diff
@@ -23,7 +23,26 @@ export default class FilterComponent<S = Signature> extends Component<S> {
...
+      console.log(params);
+
+      params.map((param) => {
+        if (queryParams[param]) {
+          values.push(queryParams[param] as string);
+        }
+      });
+      console.log(values);
+      return values.join(',');
+    }
```
aliokan, line 42: **Blocker**: You test if `queryParam` contains "," twice. One time here and one time in [select-multiple-filter.ts](https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/176/files#diff-b4e1d8e75f9d4227dd3d930f4b58770405fac4cd0afbcd478e8b3f7370168277) inserted. 

**Proposale**: 
Revert `getQueryParam` modifications, and manage the multiple `queryParam` only in [select-multiple-filter.ts](https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/176/files#diff-b4e1d8e75f9d4227dd3d930f4b58770405fac4cd0afbcd478e8b3f7370168277) and call `getQueryParam` for each param.

app/components/filters/select-multiple-filter.ts
```diff
@@ -1,29 +1,58 @@
...
-  /** Action to parse the queryParameter value(s) & select them in the select input */
   @action
   async inserted() {
-    // The queryParam is an array of searchField values, joined by seperator
-    // On the page load, we can use this to load in values
-    const queryParam = this.getQueryParam(this.args.queryParam);
-    if (queryParam) {
+    if (this.args.queryParam?.includes(',')) {
+      const queryParams = this.getQueryParam(this.args.queryParam) as string;
+      const needles = queryParams.split(/[+,]/);
```
aliokan, line 27: **Question**: When is `queryParams` contains ","?

app/components/filters/select-multiple-filter.ts
```diff
@@ -46,14 +75,35 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
...
+    if (this.args.queryParam.includes(',')) {
+      const queryParams: Record<string, unknown> = {};
+
+      selectedOptions.forEach((value) => {
+        if (queryParams[value['type'] as string]) {
+          queryParams[value['type'] as string] += '+' + value['label'];
+        } else {
+          queryParams[value['type'] as string] = value['label'];
+        }
+      });
```
aliokan, line 98: **Nitpick**: Is it more readable with `reduce`? 

```suggestion
      const queryParams = selectedOptions.reduce((acc, { label, type }) => {
        return {
          ...acc,
          ...(type && { [type]: acc[type] ? `${acc[type]}+${label}` : label }),
        };
      }, {} as Record<string, unknown>);
```

aliokan (changes requested):
@snenenenenenene I left a few comments. 

I also notice a bug when you navigate between home and list pages, the state of the component is not correct. 
1. Open home page
2. Select a municiaplity
3. Click on "Bekijk agendapunten"
4. Remove the municipality
5. Click on "Lokaal Beslist" in the header
6. Result the removed municipality is back.

app/components/filters/select-multiple-filter.ts
snenenenenenene: I've made it so that you can add multiple query params delimited by a comma (,) in the multiple select

app/components/filters/select-multiple-filter.ts
snenenenenenene: It doesn't crash if it is undefined, am I supposed to still add a handler for it then?

config/environment.js
```diff
@@ -19,6 +19,8 @@ module.exports = function (environment) {
     APP: {
       // Here you can pass flags/options to your application instance
       // when it is created
+      municipalities: 'gemeentes',
+      provinces: 'provincies',
```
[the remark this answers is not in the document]
snenenenenenene: I felt it was more readable to do it this way as I really dislike seeing Dutch words used as hard-coded strings

app/services/province-list.ts
```diff
@@ -0,0 +1,54 @@
...
+      .map(({ id }) => id);
+
+    return provinceIds.join(',');
+  }
+
+  private async _loadProvinces() {
+    const provinces = await this.store.query('administrative-unit', {
+      filter: { classification: 'Provincie' },
+      page: { size: 100 },
+    });
```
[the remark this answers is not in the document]
snenenenenenene: Doesn't it use more computing power the more filters I put on the query?

snenenenenenene:
@aliokan I've fixed the bug!

config/environment.js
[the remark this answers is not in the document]
aliokan: I agree, having "Dutch words" spread in the source code is something that bother me for a while. 

But I'm not sure that using environment variables is the best option. If we move all Dutch there it will turn messy. Moreover, ENV variables are primarily made to configure application from the outside, these queryParams are not plan to be different based on environments.  

I would go into something more simple like a constants folder: `constants/query-params.ts`

app/services/province-list.ts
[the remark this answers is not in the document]
snenenenenenene: Fixed it nowwwwwww

config/environment.js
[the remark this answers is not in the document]
snenenenenenene: On it!

app/services/province-list.ts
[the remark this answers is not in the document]
aliokan: I'm not sure about the computing power, but it adds complexity. We mix location for 'Gemeente' and administrative unit for 'Provincie'. Also, we can have multiple administrative units for one province. I would definitely use location for both.

app/components/filters/select-multiple-filter.ts
aliokan: I'm not sure to understand, isn't it what the `+` is use for? Do you have an example?

app/components/filters/select-multiple-filter.ts
aliokan: Sorry, it is not when `this.args.queryParam` but when `this.getQueryParam(this.args.queryParam)`return `undefined`. The previous version of the code was handling this case. you can simply add a guard like `if(!queryParam) return [];

aliokan (commented):
Added some comments :ping_pong:

app/components/filters/select-multiple-filter.ts
snenenenenenene: Nevermind! I am always a bit confused when talking about query params, I thought this.args.queryParam was the variable instead of the param name. I've changed the split to deserialize now!

snenenenenenene:
Right back at you @aliokan ! 🏓

LBONTEN (commented):
Can't currently test this PR since there are no agenda items attached to a province in the DEV environment. And the QA environment is currently down. Best to try again once QA is back

aliokan (commented):
Code change look good. I haven't test locally the latest version. 

LGTM in order to move forward

PR #182, opened 2024-01-05
feat: ✨ service-widget

snenenenenenene:


ember-cli-build.js
```diff
@@ -12,6 +12,8 @@ module.exports = function (defaults) {
     },
   });
 
+  app.import('vendor/jsdWidgetScript.js');
```
nvdk, line 15: I doubt this is required if you're sourcing the script directly from atlassian.com in index.html. Either include the js as packaged here in index.html or drop this script.

snenenenenenene: What do you mean exactly @nvdk ?

snenenenenenene: Oh! I must've forgotten to rebuild when I made changes to the build script because I thought it stopped working when I removed either one.

PR #185, opened 2024-01-11
Use resolution title as fallback in titleFormatted

aliokan:
## Before

image

## After

image


snenenenenenene (approved):
Very good very nice! :)

PR #186, opened 2024-01-16
fix: :bug: open accordion when a bestuursorgaan is being queried

snenenenenenene:
# BNB-490

## What?
I've made it so that when a user picks a governing body classification and returns they will still see an open advanced filter accordion. This also works when navigating to the page through a link.

## Why?
This change is more in line with the expected behaviour comparable to how searching on a title or keyword goes. See [BNB-490](https://binnenland.atlassian.net/browse/BNB-490?atlOrigin=eyJpIjoiOTdmY2M5NTg5YmMwNDM4ODhlNmE3ODczZjFkYjdhM2YiLCJwIjoiaiJ9) for more information.

## How?
As the current iteration of Appuniverse does not support a way to open an accordion through an argument.
This means that I had to create a seperate `<Accordion/>` component based off of the `<AuAccordion/>` but with an `@defaultOpen={{bool}}` argument to set the default state of the accordion.

## Testing?
I've added coverage to whether or not content gets rendered inside of the accordion component when `@defaultOpen={{bool}}` is set to either true/false.

## Screenshots
**Before**
Screenshot 2024-01-17 at 3 35 04 PM

**After**
Screenshot 2024-01-17 at 3 35 11 PM

### DIFF
--- a/app/components/accordion.hbs
+++ b/app/components/accordion.hbs
@@ -1,0 +1,54 @@
+<div class="au-c-accordion {{this.skin}} {{this.reverse}}" ...attributes>
+  <AuToolbar
+    @nowrap={{true}}
+    @reverse={{@reverse}}
+    {{on "click" this.toggleAccordion}}
+    data-test-accordion-toggle
+    as |Group|
+  >
+    <Group>
+      <div>
+        <AuButton
+          @skin="link"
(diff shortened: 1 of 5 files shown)


aliokan (commented):
Why do we need a new component?

snenenenenenene:
@aliokan I've added a proper explanation of why I did what I did

PR #191, opened 2024-01-30
feat: :sparkles: add governing body name to agenda item

snenenenenenene:


aliokan (commented):
In the PR description and the ticket, it refers to governingBody classification name like on the list page. But in the change I see governingBody name used. Is it a mistake?

snenenenenenene:
@aliokan this was indeed a mistake on my part, sorry!
i've fixed it now

aliokan (approved):
looks good! :+1:

PR #192, opened 2024-01-30
refactor: :recycle: switched to computed property for showAdvancedFilters

snenenenenenene:


app/controllers/agenda-items/index.ts
```diff
@@ -171,7 +172,10 @@ export default class AgendaItemsIndexController extends Controller {
   @tracked plannedStartMax = '';
   @tracked governingBodyClassifications = '';
 
-  @tracked showAdvancedFilters = this.governingBodyClassifications.length > 0;
+  @computed('governingBodyClassifications.length')
```
aliokan, line 175: **Blocker:** Using a getter is enough, `@computed` decorator is not needed. 
In the previous version, `this.governingBodyClassifications.length > 0` was only evaluate and pass to `showAdvancedFilters` only once at initialization. With the getter, the component will know and refresh each time `governingBodyClassifications` change. cf. https://guides.emberjs.com/v4.12.0/in-depth-topics/autotracking-in-depth/

aliokan (changes requested):

app/controllers/agenda-items/index.ts
```diff
@@ -1,4 +1,5 @@
 import Controller from '@ember/controller';
+// eslint-disable-next-line ember/no-computed-properties-in-native-classes
```
aliokan, line 2: **Nitpick:** Can you remove that line too :smile_cat:

aliokan (changes requested):
One little change and we are good to go! :ping_pong:

aliokan (approved):
LGTM!

PR #193, opened 2024-01-31
fix: 🐛 antwerpen-query-select-fix

snenenenenenene:
# BNB-524

## What?
There was a big bug inside of the `inserted` function inside of the select-filter component. In fact, there were **2 big issues** occuring either when _being routed from home to the agenda items route_ or _when being directed to the agenda items route through a link_:
1. Whenever multiple governments would get queried they wouldn't get delimited from each other. This in turn made it so that if a user were to select for instance "Aalst+Aarschot" the values wouldn't get processed properly by the haystack/needle function as it would be looking for "Aalst+Aarschot" instead of the "Aalst" and "Aarschot".
2. Whenever the haystack/needle function would look for a government it would do so regardless of whether it is a `province` or  a `municipality`- it would always find the `municipality` first because this occurs before the provinces in the haystack. This in turn means that if a user would query a province with a municipality counterpart (Antwerp) it would always select the municipality version of Antwerp which then made it select unable to properly communicate with the actual query params.

## Why?
Previously if a user was to select multiple governments and/or the province of Antwerp and then continue to the agenda items route there would be a bug where the multiple governments would not show up in the select component and/or the province of Antwerp would not get removed from the query parameters through said select component. 
[BNB-524](https://binnenland.atlassian.net/browse/BNB-524) for more information.

## How?
As the current iteration of Appuniverse does not support a way to open an accordion through an argument.
This means that I had to create a seperate `<Accordion/>` component based off of the `<AuAccordion/>` but with an `@defaultOpen={{bool}}` argument to set the default state of the accordion.

## Testing?
Tests should also be written for this improved functionality, but I reckon we should be creating a seperate and more congruent test PR for the filter components.

## Screenshots
**Before**
https://github.com/lblod/frontend-burgernabije-besluitendatabank/assets/50323795/6e524bf2-748d-4d2b-a049-feef21e0f92a

**After**

https://github.com/lblod/frontend-burgernabije-besluitendatabank/assets/50323795/7bd5f6b2-f7df-4a16-997f-7adc92fe1090

### DIFF
--- a/app/components/filters/select-multiple-filter.ts
+++ b/app/components/filters/select-multiple-filter.ts
@@ -33,7 +33,10 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
       const needles = deserializeArray(this.args.queryParam).map(
         (queryParam) => {
           if (!queryParam) return [];
-          return this.getQueryParam(queryParam);
+          return {
+            queryParam: queryParam,
+            value: this.getQueryParam(queryParam),
+          };
         }
       );
 
@@ -52,11 +55,28 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
       }
 
       needles.forEach((needle) => {
-        const found = flattenedHaystack.find(
-          (value) => get(value, searchField) === needle
-        );
-        if (found) {
-          results.push(found);
+        if (
+          typeof needle === 'object' &&
+          'value' in needle &&
+          typeof needle.value === 'string'
+        ) {
+          const splitNeedles = needle.value.split('+').map((value) => {
+            return {
+              queryParam: needle.queryParam,
+              value: value,
+            };
+          });
+
+          splitNeedles.forEach((needle) => {
+            const found = flattenedHaystack.find(
+              (value) =>
+                get(value, searchField) === needle.value &&
+                value['type'] === needle.queryParam
+            );
+            if (found) {
+              results.push(found);
+            }
+          });
         }
       });
 
--- a/app/components/filters/select-multiple-filter.ts
+++ b/app/components/filters/select-multiple-filter.ts
@@ -1,4 +1,6 @@
 import { action, get } from '@ember/object';
+import RouterService from '@ember/routing/router-service';
+import { inject as service } from '@ember/service';
 import { deserializeArray } from 'frontend-burgernabije-besluitendatabank/utils/query-params';
 import FilterComponent, { type FilterArgs } from './filter';
 
@@ -18,6 +20,8 @@ interface Signature {
 }
 
 export default class SelectMultipleFilterComponent extends FilterComponent<Signature> {
+  @service declare router: RouterService;
+
   get selected() {
     return this.args.selected;
   }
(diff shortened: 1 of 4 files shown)


aliokan (changes requested):
@snenenenenenene Can you add some tests to validate the fix?

app/components/filters/select-multiple-filter.ts
```diff
@@ -52,11 +55,28 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
...
       needles.forEach((needle) => {
-        const found = flattenedHaystack.find(
-          (value) => get(value, searchField) === needle
-        );
-        if (found) {
-          results.push(found);
+        if (
+          typeof needle === 'object' &&
+          'value' in needle &&
+          typeof needle.value === 'string'
```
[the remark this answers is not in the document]
snenenenenenene: @aliokan it should be better now!

snenenenenenene:
@aliokan I've processed the feedback!

app/components/filters/select-multiple-filter.ts
```diff
@@ -29,59 +33,61 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
...
-      const searchField = this.args.searchField;
+    if (!this.router.currentRoute) {
+      console.error('Current route is not available');
+      return;
+    }
 
-      const haystack = (await this.args.options) as unknown as GroupedOptions[];
-      const results: Option[] = [];
+    const results: Option[] = [];
+    const haystack = (await this.args.options) as unknown as GroupedOptions[];
```
aliokan, line 44: **Question:** Why do we need to override the type here. Can't we just fix the type in Signature?

https://github.com/lblod/frontend-burgernabije-besluitendatabank/blob/945b617459978a3bb578a7a4f2081113d74116c1/app/components/filters/select-multiple-filter.ts#L14-L20

app/components/filters/select-multiple-filter.ts
```diff
@@ -29,59 +33,61 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
...
+          (value) =>
+            get(value, searchField) === needle.value &&
+            value['type'] === needle.queryParam
         );
         if (found) {
           results.push(found);
         }
-      });
+      }
+    });
```
aliokan, line 88: **Suggestion:** I would like to go further on the simplification by merging the `needles` and `results` loops. 

What do you think about this example?  
```typescript
const results = deserializeArray(this.args.queryParam).flatMap(
  (queryParam) => {
    if (!queryParam) return [];
    const queryParamValue = this.getQueryParam(queryParam);
    const values = queryParamValue ? deserializeArray(queryParamValue) : [];
    return values
      .map((value) => {
        return flattenedHaystack.find(
          (option) =>
            get(option, searchField) === value &&
            option['type'] === queryParam
        );
      })
      .filter(Boolean) as Option[];
  }
);
```

aliokan (commented):
I left suggestion in the `inserted` function. :ping_pong: 
And thank you for the integration tests! :pray:

app/components/filters/select-multiple-filter.ts
snenenenenenene: @aliokan because if it gets to that point in the function the type of the `options` array is no longer an array of `Record<string, string>` but instead it is now in the format of powerselect's 'grouped options'.
e.g. `options: [
  groupName: string;
  options: Option[];
]
` instead of `options: Option[]`

aliokan: This is a design error. We cannot declare one type and use another one. 
It not directly related to this bug fix, but we should clean that in a follow up (BNB-546).

aliokan (approved):
Looks good! :+1:

PR #196, opened 2024-02-15
Fix governing body filter inital state

aliokan:
Related to BNB-555

### DIFF
--- a/app/components/filters/select-multiple-filter.ts
+++ b/app/components/filters/select-multiple-filter.ts
@@ -11,9 +11,11 @@ type GroupedOptions = {
   options: Option[];
 };
 
+type UnifiedOptions = Option | GroupedOptions;
+
 interface Signature {
   Args: {
-    options: Promise<Option[]>;
+    options: Promise<UnifiedOptions[]>;
     selected: Option[];
     updateSelected: (selected: Option[]) => void;
   } & FilterArgs;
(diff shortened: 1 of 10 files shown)


snenenenenenene:
I still found a bug when selecting multiple governing body classifications

snenenenenenene:
i've worked a small solution myself if you want to take some inspiration for this fix [https://github.com/lblod/frontend-burgernabije-besluitendatabank/tree/fix/governing-body-classification-inserted](here)

app/components/filters/select-multiple-filter.ts
```diff
@@ -33,22 +35,20 @@ export default class SelectMultipleFilterComponent extends FilterComponent<Signa
...
+    const haystack = await this.args.options;
+    const flattenedHaystack = haystack.reduce((acc, value) => {
+      if (Array.isArray(value.options)) {
+        return [...acc, ...value.options];
+      } else {
+        return [...acc, value as Option];
+      }
+    }, [] as Option[]);
 
     const results = deserializeArray(this.args.queryParam).flatMap(
```
snenenenenenene, line 53: make it so the value gets split on ",". This way when multiple classifications get selected they'll get split up properly.

```ts
const results = deserializeArray(this.args.queryParam).flatMap(
      (queryParam) => {
        if (!queryParam) return [];
        const queryParamValue = this.getQueryParam(queryParam);
        const values = queryParamValue ? deserializeArray(queryParamValue) : [];
        return values
          .flatMap((value) => value.split(','))
          .map((value) => {
            return flattenedHaystack.find(
              (option) =>
                get(option, searchField) === value &&
                option['type'] === queryParam
            );
          })
          .filter(Boolean) as Option[];
      }
    );
```

snenenenenenene (changes requested):

app/components/filters/select-multiple-filter.ts
snenenenenenene: Other than that it looks very good! :)

aliokan: @snenenenenenene The problem was that we mix serialization separator. We used `+` on locations and `,` on governingBodyClassification. I refactor the `selectChange` method in order to apply `+` everywhere, that fix the bug too.

snenenenenenene (approved):
LGTM!

PR #204, opened 2024-05-02
Add a way to disable governing body via env variable

aliokan:
Related to BNB-617

Add a way to disable governing bodies through environment variable. 

Agenda-item linked to disabled governing body are not visible on the list page, and detail page are redirected to the list page.


snenenenenenene (approved):
LGTM!

PR #205, opened 2024-11-05
BNB-574 & BNB-573 Adapt the jira iframe to the screen height and tran…

Andresdev02:
BNB-574 & BNB-573 Adapt the jira iframe to the screen height and translate the text into dutch

- Replace text "Contact us about" to Dutch translation
- Make pop-up not bigger then the screen height

I also removed the footer not sure if this is expected or not. I can add it back again in the future.


brechtvdv (approved):
Looks good

PR #207, opened 2025-01-02
Add max height to container to it will be able to scroll on mobile

Andresdev02:


brechtvdv (approved):
Works on different mobile phones

PR #209, opened 2025-01-15
Feature/update ember and ember data to the new lts

Andresdev02:


ember-cli-build.js
Windvis, line 1: Hmm, ideally we would keep using Embroider here.

The newer ember blueprints now have an `--embroider` flag, which gives us updates to the Embroider setup. If we added that flag to [this config file](https://github.com/lblod/frontend-burgernabije-besluitendatabank/blob/4c63bf6375686246834bce5d341fd0c3d00465fb/config/ember-cli-update.json#L13-L15) before running ember-cli-update, this wouldn't have been reverted, but I forgot about that, sorry.

Since the blueprint is already updated we need a different way of switching to that variant of the blueprint.

I think you can run `npx ember init --embroider --typescript` and then do some manual git diff editing to get the result you want. We most likely  need changes to the ember-cli-build.js, package.json (embroider has slightly different dependencies) and the ember-cli-update config file so the option is added there.

This might be a bit tricky, so ping me if you need help.

ember-cli-build.js
```diff
@@ -1,49 +1,21 @@
...
+      cascade: true,
+      sourcemap: true,
     },
-  };
-
-  if (isProductionBuild) {
-    // `staticEmberSource` breaks the Ember dev tools: https://github.com/embroider-build/embroider/issues/1575
-    // As a workaround we only enable it for production builds so that we can still use the dev tools for local development
-    // while still having a smaller bundle in production.
-    embroiderOptions.staticEmberSource = true;
```
Windvis, line 41: FYI, this is no longer the case, so this condition is no longer needed. (but the npx ember init should do that for you).

ember-cli-build.js
```diff
@@ -1,49 +1,21 @@
...
-    packagerOptions: {
-      webpackConfig: {
-        plugins: [
-          process.env.ANALYZE_BUNDLE ? new BundleAnalyzerPlugin() : null,
-        ].filter(Boolean),
-      },
+    autoprefixer: {
+      enabled: true,
+      cascade: true,
+      sourcemap: true,
```
Windvis, line 16: Did you copy this from a different app? I don't see this in the original file and I don't think it's part of the blueprint?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frontend-burgernabije-besluitendatabank",
-  "version": "0.8.1",
+  "version": "0.8.2",
```
Windvis, line 3: Why was this bumped here? Bumping in branches isn't a good idea :smile: If you want an image to test before merging I would use the "feature" image builds.

package.json
```diff
@@ -34,100 +34,105 @@
...
-    "ember-page-title": "^7.0.0",
-    "ember-plausible": "^0.2.0",
-    "ember-power-select": "^7.1.0",
-    "ember-qunit": "^6.2.0",
-    "ember-resolver": "^10.0.0",
-    "ember-resources": "^6.3.1",
-    "ember-source": "~4.12.0",
-    "ember-template-lint": "^5.7.2",
-    "ember-test-selectors": "^6.0.0",
+    "ember-cli-typescript": "^5.3.0",
```
Windvis, line 95: I think ember-cli-typescript is no longer needed.

package.json
```diff
@@ -34,100 +34,105 @@
...
     "broccoli-asset-rev": "^3.0.0",
-    "concurrently": "^8.0.1",
-    "date-fns": "^2.29.3",
-    "ember-auto-import": "^2.6.3",
-    "ember-cli": "~4.12.1",
-    "ember-cli-app-version": "^6.0.0",
+    "concurrently": "^8.2.2",
+    "date-fns": "^4.1.0",
+    "ember-auto-import": "^2.10.0",
+    "ember-cli": "~6.1.0",
```
Windvis, line 87: Strange, did you bump this to 6.1 yourself? Ideally this matches the ember-source version (unless there is a good reason not to).

package.json
```diff
@@ -34,100 +34,105 @@
...
-    "ember-power-select": "^7.1.0",
-    "ember-qunit": "^6.2.0",
-    "ember-resolver": "^10.0.0",
-    "ember-resources": "^6.3.1",
-    "ember-source": "~4.12.0",
-    "ember-template-lint": "^5.7.2",
-    "ember-test-selectors": "^6.0.0",
+    "ember-cli-typescript": "^5.3.0",
+    "ember-concurrency": "^3.1.1",
+    "ember-data": "~4.12.8",
```
Windvis, line 97: This should also have been updated to v5+, I think v5.3.9 is the latest stable version at the moment.

Windvis (commented):
Some feedback after looking at the code :+1:

app/services/store.ts
Windvis, line 1: Due to a weird quirk in Ember this has to be a .js file. At the moment the .js file from EmberData still "wins" and this one isn't being used (so the deprecation is still shown). I think this would solve itself eventually once EmberData removes its own store.js file, but we would see the message until that happens, so I would just rename to .js for now, especially since it's only a reexport.

app/routes/agenda-items/session.ts
```diff
@@ -15,14 +15,14 @@ export default class AgendaItemsAgendaItemSessionRoute extends Route {
 
     const session = await this.store.findRecord(
       'session',
-      sessions?.firstObject?.id as string,
+      sessions?.objectAt(0)?.id as string,
```
Windvis, line 18: `.objectAt` is an EmberArray thing which doesn't exist here anymore (I think EmberData v5 removed that, but it should have shown a deprecation in v4 as well).

```suggestion
      sessions?.at(0)?.id as string,
```

package.json
```diff
@@ -34,100 +34,106 @@
...
+    "ember-concurrency": "^3.1.1",
+    "ember-data": "~5.3.9",
+    "ember-focus-trap": "^1.1.1",
+    "ember-load-initializers": "^3.0.1",
+    "ember-modifier": "^4.2.0",
+    "ember-modify-based-class-resource": "^1.1.1",
+    "ember-page-title": "^8.2.3",
+    "ember-plausible": "^0.4.0",
+    "ember-power-select": "^7.2.0",
+    "ember-qunit": "^9.0.1",
```
Windvis, line 105: Hmm, I think this might be the cause of the test faillure in CI.

I'm guessing you bumped this to v9 manually? I think you either need to downgrade to v8, or do something like this to make it work in v9: https://github.com/ember-cli/ember-cli/commit/898ecfec75f804aa3dedf8bc33aa011b6087028e

Windvis:
The feature build is failing because the Dockerfile needs to be updated to node 22 as well. The docker-ember image that's used there still uses v16 internally.

Windvis:
Still missing this part of https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/209#issuecomment-2594958468 :sweat_smile: 

> Also, the build command in the dockerfile needs to be updated to something like this: https://github.com/lblod/frontend-toezicht-abb/blob/4d2a5e7ae5c497db62ffe2befdb2d79c20345237/Dockerfile#L9

> The node images don't have ember-cli installed globally.

Other than that, it looks good to me :+1: . I did a brief test, but I don't know the app so well.

Andresdev02:
@Windvis thank you for the good review! It should be ok now. 😅

Windvis (approved):

ember-cli-build.js
```diff
@@ -1,49 +1,26 @@
...
-    staticComponents: true,
-    staticEmberSource: isProductionBuild,
-
-    // This config is needed to work around an Ember Data v4.12 + Embroider issue
-    // We can remove it once the fix is released: // https://github.com/embroider-build/embroider/issues/1506
-    compatAdapters: new Map([['@ember-data/debug', null]]),
-    packagerOptions: {
-      webpackConfig: {
-        plugins: [
-          process.env.ANALYZE_BUNDLE ? new BundleAnalyzerPlugin() : null,
```
Windvis, line 31: Whoops, one last thing. The bundle analyzer is no longer used now, so I would remove the dependency and `build:analyze-bundles` script in the package.json file.

PR #214, opened 2025-01-28
Add new status filter

Andresdev02:
BNB-329

Filter on whether or not an agenda item has been treated or not

This can be tested by running the frontend: 
`ember s --proxy https://lokaalbeslist.lblod.info/`


brechtvdv (approved):
Works fine

PR #216, opened 2025-01-30
Feature/bnb 316 add nog niet behandeld explanations

Andresdev02:
Add `nog niet behandeld` explanations

Currently, when the data for a decision or vote is unavailable on the agenda detail page, we display a generic "Data niet beschikbaar" message. This message does not provide sufficient context to the user, as there are multiple reasons why the decision or vote data might not be available (e.g., data not harvested yet, or the agenda item has not yet been handled).

We want to improve the user experience by replacing the generic message with a contextual alert-card that includes a specific title and explanatory body text based on the reason for the missing data.

Acceptance Criteria:

Replace the current "Data niet beschikbaar" message with an alert-card component.

The alert-card should display different titles and body text based on the specific reason for the missing data:

Case 1: Data not harvested yet

Title: Besluit/Stemming Niet Beschikbaar

Body: De data voor dit agendapunt is nog niet beschikbaar.

Case 2: Agenda item not yet handled

Title: Besluit/Stemming Niet Beschikbaar

Body: Dit agendapunt werd nog niet behandeld.

The logic to determine the reason should be based on the status of the agenda point (e.g., harvested status, handled status).

Test Scenario:

When the data for an agenda point is not harvested,
If the detail page is loaded,
Then the alert-card should display:

Title: Besluit/Stemming Niet Beschikbaar

Body: De data voor dit agendapunt is nog niet beschikbaar.

When the agenda point has not been handled yet,
If the detail page is loaded,
Then the alert-card should display:

Title: Besluit/Stemming Niet Beschikbaar

Body: Dit agendapunt werd nog niet behandeld.

When the decision or vote data is available for an agenda point,
If the detail page is loaded,
Then the alert-card should not be displayed.

This can be tested by running the frontend:
`ember s --proxy https://lokaalbeslist.lblod.info/`

### DIFF
--- a/app/components/vote-overview.ts
+++ b/app/components/vote-overview.ts
@@ -32,11 +32,15 @@ export default class VoteOverview extends Component<ArgsInterface> {
   }
 
   get hasVotersData() {
-    return (
-      this.args.vote.hasAbstainers.slice().length > 0 ||
-      this.args.vote.hasOpponents.slice().length > 0 ||
-      this.args.vote.hasProponents.slice().length > 0
-    );
+    const abstainers = this.args.vote.hasAbstainers;
+    const opponents = this.args.vote.hasOpponents;
+    const proponents = this.args.vote.hasProponents;
+
+    const hasAbstainers = Array.isArray(abstainers) && abstainers.length > 0;
+    const hasOpponents = Array.isArray(opponents) && opponents.length > 0;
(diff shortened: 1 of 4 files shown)


brechtvdv (commented):
I added a comment about Articles in Jira https://binnenland.atlassian.net/browse/BNB-316

brechtvdv (approved):

PR #225, opened 2025-03-19
Postpone loading of related agendaitems for faster first load of agenda-item

madnificent:
The first load is quite slow now because of loading of related agenda-points in the detail view.  Database fine tuning will likely help but rendering async will buy us some time to figure out the best approach to that.


Andresdev02 (approved):
Tested and loading a lot faster now. 🚀

